Issue #62, opened 2026-05-03
Architecture: separate the agent core library from the CLI adapter


zjshen14:
## Background

The agent and CLI are structurally coupled today. `createAgent()` in `src/cli/index.ts` mixes library concerns (constructing the LLM client, assembling the tool registry) with CLI concerns (reading `~/.opencli/config.json`, loading API keys from env). `runRepl()` in `src/cli/repl.ts` mixes UI concerns with session persistence side effects.

As a result:
- The agent cannot be used as a Node.js library without pulling in CLI dependencies
- There is no path to a web API front-end, a programmatic test harness, or an MCP server adapter without significant refactoring
- Unit testing requires mocking terminal I/O and the filesystem config

## Proposed package boundary

Introduce a strict source-level boundary between the library core and the CLI adapter:

```
src/
  core/           ← pure library — no process.env, no file I/O, no terminal deps
    agent.ts      (Agent class)
    executor.ts
    context.ts
    prompt.ts
  providers/      ← LLM clients — no CLI deps
    gemini.ts
    anthropic.ts
    openai.ts
    factory.ts
    schema.ts
    types.ts
  tools/          ← tool implementations — no CLI deps
    ...
  skills/         ← skill loading — no CLI deps
    ...
  state/          ← config and session I/O (used by CLI, not by core)
    config.ts
    session.ts
  cli/            ← thin adapter: reads config, wires dependencies, runs REPL
    index.ts
    repl.ts
    renderer.ts
    input.ts
```

The rule: `core/` and `providers/` and `tools/` import nothing from `cli/` or `state/`. The agent constructor takes all dependencies explicitly — no `loadConfig()` calls, no `process.env` reads inside the core.

```ts
// programmatic usage becomes possible:
import { Agent } from "opencli/core";
import { AnthropicClient } from "opencli/providers";
import { createDefaultRegistry } from "opencli/tools";

const agent = new Agent(
  new AnthropicClient(apiKey, "claude-opus-4-7"),
  createDefaultRegistry(),
  skills,
  systemInstruction,
);

for await (const event of agent.run("refactor the auth module")) {
  // handle events
}
```

The CLI adapter (`src/cli/`) is the only layer that reads config files and environment variables, then passes values down as constructor arguments.

## Acceptance criteria

- [ ] `src/core/`, `src/providers/`, `src/tools/` contain no imports from `src/cli/` or `src/state/`
[...]

PR #66, opened 2026-05-03
refactor: separate library core from CLI adapter

zjshen14:
## Summary

Implements #62 — enforces a clean boundary between the library core and the CLI adapter so the agent can be used programmatically without importing any CLI modules.

- **`src/agent/` → `src/core/`** (`agent/core.ts` renamed to `core/agent.ts` to avoid redundancy)
- **`src/model/` → `src/providers/`** (all provider clients and types)
- **`providers/factory.ts`**: removed `Config` import and `process.env` reads — `createClient(model, apiKey)` now accepts an explicit key
- **`cli/index.ts`**: new `resolveApiKey()` function owns all env-var + config-file resolution; the resolved key is passed down to `createClient()` — the library layer never reads `process.env` or config files
- **CLAUDE.md / AGENTS.md**: source structure, architecture description, and import boundary rule updated

## The boundary rule

> `core/` and `providers/` must never import from `cli/` or `state/`.

Verified by grep — zero `../cli/` or `../state/` imports remain in either directory.

## Programmatic usage (now works)

```ts
import { Agent } from "./src/core/agent.js";
import { AnthropicClient } from "./src/providers/anthropic.js";
import { createDefaultRegistry } from "./src/tools/index.js";
import { SkillRegistry } from "./src/skills/registry.js";

const client = new AnthropicClient(process.env.ANTHROPIC_API_KEY!, "claude-opus-4-7");
const agent = new Agent(client, createDefaultRegistry(), new SkillRegistry(), systemPrompt);

for await (const event of agent.run("refactor the auth module")) {
  if (event.type === "text") process.stdout.write(event.text);
}
```

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean
- [x] `npm run format:check` — clean
- [x] `npm test` — 307 tests, all pass
- [x] Pre-commit hook ran all checks on commit


zjshen14:
## Code Review

**Status: Approve with minor comments**

---

### What's Good

- **Boundary is correctly drawn.** `factory.ts` drops to 9 lines once `process.env` and `Config` are removed — that simplification validates the design.
- **`resolveApiKey()` placement is right.** Env vars + config file access belong in the CLI layer; the function signature `(provider, config) → string` is clean and testable.
- **Test files moved alongside source.** Follows the colocation convention in `CLAUDE.md`.
- **Docs updated in sync.** `CLAUDE.md` and `AGENTS.md` both updated in the same commit.

---

### Issues

**1. `resolveApiKey()` has an implicit Gemini fallthrough**

`src/cli/index.ts` — the function returns Gemini as a default for any non-anthropic/non-openai provider. If a new provider is added to the `Provider` type but not handled here, it silently gets a Gemini key and likely produces a confusing runtime error downstream. A narrowed `else` with an exhaustiveness check would close this:

```ts
// after the openai branch:
if (provider !== "gemini") {
  const _: never = provider;
  throw new Error(`Unknown provider: ${provider}`);
}
const key = process.env.GEMINI_API_KEY ?? config.geminiApiKey;
```

**2. `resolveApiKey()` is untested**

It's a pure function over `process.env` + `Config` — straightforward to unit test with a few key-present/key-absent cases. This would also catch the implicit-fallthrough issue above.

**3. `factory.test.ts` worth a quick scan**

The test file was only renamed in the diff. Worth confirming the tests exercise the new `(model, apiKey: string)` signature rather than the old `(model, config)` one.

---

### Minor Suggestions (non-blocking)

- **No barrel exports.** The programmatic usage example imports from internal paths (`./src/core/agent.js`). A `src/index.ts` exporting `{ Agent, createClient, ... }` would be the right follow-up before advertising library usage externally.
- **Boundary is not mechanically enforced.** An ESLint `no-restricted-imports` rule on `src/core/` and `src/providers/` would prevent regressions automatically. Low priority but worth a follow-up issue.

---

### Summary

Low-risk, correctly executed rename refactor. The main concern worth addressing before merge is the implicit Gemini fallthrough in `resolveApiKey()` — it's a silent correctness hazard when the `Provider` union grows. Everything else is non-blocking.

zjshen14:
Addressed all three review points in commit `08ceff9`:

**1. Exhaustiveness guard on `resolveApiKey`**
Replaced the implicit Gemini fallthrough with an explicit `if (provider === "gemini")` branch, followed by a `never` guard. TypeScript will now error at compile time if a new `Provider` value is added without handling it here.

**2. `resolveApiKey` extracted and tested**
Moved to `src/cli/keys.ts` (necessary since `cli/index.ts` can't be imported in tests — `program.parseAsync` fires at module load). Added `src/cli/keys.test.ts` with 10 tests covering:
- env var present → returned
- config key present → returned  
- env var takes precedence over config
- neither present → throws with correct message
…for all three providers (anthropic, openai, gemini).

**3. `factory.test.ts`**
Confirmed it only tests `detectProvider` and `hasNativeThinking`, which have unchanged signatures. `createClient` itself is a thin dispatcher (3 lines) that's exercised by the integration tests.

317 tests, all checks clean.

Issue #44, opened 2026-05-03
Bug: encodeProjectPath collides on hyphenated directory names


zjshen14:
## Problem

`encodeProjectPath` in `src/state/session.ts:21` converts the working directory path to a storage key by replacing `/` with `-`:

```ts
return cwd.replace(/\//g, "-");
```

`/home/alice/project` and `/home/alice-project` both encode to `-home-alice-project`. On any system with hyphenated directory names this silently merges two unrelated projects into the same session log directory — new sessions from either project append to the same JSONL files, and `Session.list()` returns interleaved history.

## Expected behavior

Each unique `cwd` maps to a unique storage directory. No two distinct paths should produce the same key.

## Fix

Replace `/` with a separator that cannot appear in a POSIX path, or hash the path:

```ts
// Option A — base64url (lossless, no collisions)
function encodeProjectPath(cwd: string): string {
  return Buffer.from(cwd).toString("base64url");
}

// Option B — URL-encode (human-readable-ish)
function encodeProjectPath(cwd: string): string {
  return encodeURIComponent(cwd).replace(/%2F/gi, "~");
}
```

Note: a migration path is needed for users who already have session data stored under the old encoding.

## Location

`src/state/session.ts:21` — `encodeProjectPath()`

Issue #46, opened 2026-05-03
Bug: session ID has no uniqueness within the same second — concurrent starts corrupt logs


zjshen14:
## Problem

Session IDs are generated as `YYYY-MM-DDTHH-mm-ss` in `src/state/session.ts:25`:

```ts
function makeSessionId(): string {
  return new Date().toISOString().slice(0, 19).replace(/:/g, "-");
}
```

Two sessions started within the same second (e.g. a script that spawns agents, or a fast test suite) produce the same ID. Because sessions are stored as `<id>.jsonl`, `appendFile` silently merges both sessions into a single file. On resume, `reconstructMessages()` sees an interleaved log from two independent conversations and produces a corrupt `Message[]` with no error.

## Expected behavior

Every `Session.create()` call produces a globally unique ID.

## Fix

Append milliseconds or a short random suffix:

```ts
function makeSessionId(): string {
  const ts = new Date().toISOString().slice(0, 23).replace(/[:.]/g, "-");
  return ts; // millisecond precision is enough for human starts
}

// Or, if sub-millisecond safety is needed:
function makeSessionId(): string {
  const ts = new Date().toISOString().slice(0, 19).replace(/:/g, "-");
  return `${ts}-${Math.random().toString(36).slice(2, 6)}`;
}
```

## Location

`src/state/session.ts:25` — `makeSessionId()`

PR #67, opened 2026-05-03
Fix session ID and encodeProjectPath collisions

zjshen14:
This PR fixes issue #44 and #46 regarding session and project path collisions.

- Session IDs now have millisecond precision.
- Project paths are URL-encoded to avoid collisions with hyphens.
- A seamless migration script is included to move old directories to the new format if they exist.

Co-authored-by: Antigravity <antigravity@google.com>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1489,6 +1489,7 @@
       "integrity": "sha512-rLoGZIf9afaRBYsPUMtvkDWykwXwUPL60HebR4JgTI8mxfFe2cQTu3AGitANp4b9B2QlVru6WzjgB2IzJKiCSA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@typescript-eslint/scope-manager": "8.58.0",
         "@typescript-eslint/types": "8.58.0",
@@ -1829,6 +1830,7 @@
       "integrity": "sha512-UVJyE9MttOsBQIDKw1skb9nAwQuR5wuGD3+82K6JgJlm/Y+KI92oNsMNGZCYdDsVtRHSak0pcV5Dno5+4jh9sw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -2480,6 +2482,7 @@
       "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
+      "peer": true,
       "bin": {
         "esbuild": "bin/esbuild"
       },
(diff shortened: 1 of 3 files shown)


zjshen14:
## Review — Needs more work before merge

The two underlying bug fixes are correct (session ID millisecond precision is right, and the URL-encode approach fixes the common hyphen-collision case). But there are two issues that should be addressed first.

---

### 1. `encodeProjectPath` can still collide on paths containing `~`

`~` is in `encodeURIComponent`'s safe set, so it passes through unencoded. That means:

```
/home/alice~project  →  encodeURIComponent  →  %2Fhome%2Falice~project  →  ~home~alice~project
/home/alice/project  →  encodeURIComponent  →  %2Fhome%2Falice%2Fproject →  ~home~alice~project
```

Same encoded key — same bug as before, just for a different character. The acceptance criterion in #44 was *"No two distinct paths should produce the same key."* The `base64url` option from the issue is collision-free by construction and avoids this:

```ts
function encodeProjectPath(cwd: string): string {
  return Buffer.from(cwd).toString("base64url");
}
```

It's not human-readable but it is lossless and deterministic.

---

### 2. Silent data loss in the migration when both directories coexist

The migration only renames old → new when the new directory doesn't yet exist:

```ts
try {
  await stat(newDir);   // if this succeeds, skip the rename
} catch {
  await rename(oldDir, newDir);
}
```

If a user ran the updated binary once (creating `newDir` with fresh sessions), subsequent runs will find `newDir` already present and leave `oldDir` untouched — the old sessions are silently unreachable from `list` and `resume`. They're not deleted, but they're effectively lost from the user's perspective with no warning.

The safe fix is to merge rather than skip when both exist — move the `.jsonl` files from `oldDir` into `newDir`:

```ts
if (oldStats.isDirectory()) {
  let newExists = false;
  try { await stat(newDir); newExists = true; } catch { /* doesn't exist */ }

  if (!newExists) {
    await rename(oldDir, newDir);
  } else {
    // merge: move each old session file into the new dir
    const files = await readdir(oldDir).catch(() => []);
    await Promise.all(
      files
        .filter((f) => f.endsWith(".jsonl"))
        .map((f) => rename(join(oldDir, f), join(newDir, f)).catch(() => {})),
    );
  }
}
```

---

### 3. Minor — file-level comment is stale

The JSDoc at the top of `session.ts` still says:

```
* The project path is encoded by replacing "/" with "-".
* Session ID: YYYY-MM-DDTHH-mm-ss — … no collision risk (users can't start two sessions in the same second).
```

Both statements are no longer true. Worth updating.

---

### 4. No tests for the migration paths

Two cases aren't covered:
- old dir exists, new dir does not → rename happens
- old dir exists, new dir also exists → merge/skip (whichever is chosen)

These are the exact scenarios that could silently lose data, so they're worth testing explicitly.

---

### Summary

| | Status |
|---|---|
| Session ID millisecond precision | ✅ correct |
| URL-encode project path (common case) | ✅ fixes the reported bug |
| `~` collision edge case | ❌ needs fix |
| Migration silent data loss | ❌ needs fix |
| Migration tests | ❌ missing |
| Stale comment | minor |

Happy to approve once the `encodeProjectPath` collision and migration merge-vs-skip are addressed.

zjshen14:
All requested changes have been addressed! 

- Updated the project path encoding to use `base64url` to prevent collisions.
- Updated the migration script to correctly merge `.jsonl` files if both the old and new directories already exist.
- Added comprehensive tests for both migration scenarios.
- Updated the stale JSDoc comment.
- Included the `package-lock.json` updates.

PR #94, opened 2026-05-10
feat(cli): @file mention expansion + per-CWD input history (A4)

### LINKED ISSUE #88: feat: Phase A4 — @file mention expansion + persistent input history
## Overview

Phase A4 in the [roadmap](docs/roadmap.md): two improvements to input handling.

**@file mentions** — resolve `@path` references in user input before sending to the LLM. `@src/core/agent.ts` expands to the file contents inline, so the user doesn't have to ask the agent to read it first.

**Persistent history** — persist REPL input history per-CWD to `~/.opencli/history/<encoded-cwd>`. Today history is lost on exit.

## Scope

- `@path` expansion in `src/cli/input.ts` or `src/cli/repl.ts` — resolve before message is sent to `Agent.run()`
- Support glob patterns: `@src/**/*.ts` expands to all matching files (with a size cap)
- History persistence: load on REPL start, append on each user submit, respect `historySize` config

## Acceptance criteria

- [ ] `@src/core/agent.ts` in a prompt inlines the file content before the agent sees it
- [ ] `@nonexistent.ts` produces a clear error, does not crash
- [ ] Input history survives a REPL restart in the same CWD
- [ ] History file respects `historySize` config (prunes oldest entries)
- [ ] `npm run typecheck && npm run lint && npm test` pass

zjshen14:
## Summary

- **`@file` / `@glob` expansion** — `@src/core/agent.ts` or `@src/**/*.ts` in a prompt is resolved and inlined before the message reaches the agent. Each file is wrapped in a fenced block (`--- @path ---` / `--- end ---`). Tokens that don't resolve are warned and left unchanged (so `@user`, `@TODO` etc. are not broken).
- **Per-CWD history** — input history is now stored at `~/.opencli/history/<base64url-encoded-cwd>` instead of a single global file. Histories across projects are fully independent. Old global history is left in place untouched.

Size caps: 50 000 chars/file, 20 files/glob, 200 000 chars total per glob. Binary files (null-byte heuristic) are skipped. Original unexpanded input is saved to history (not the expanded content).

## Test plan

- [x] 9 `src/cli/mentions.test.ts` tests: single file, missing file, glob, empty glob, size cap, file-count cap, binary skip, no-@ fast path, inline expansion
- [x] 4 new `src/cli/input.test.ts` tests: per-CWD isolation, round-trip, missing file → [], cap at 500
- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — 461 tests pass

Closes #88

### DIFF
--- a/docs/design/a4-file-mentions-history.md
+++ b/docs/design/a4-file-mentions-history.md
@@ -1,6 +1,6 @@
 # Design: A4 — @file mention expansion + per-CWD input history
 
-_Status: Implemented — merged in <sha> (2026-05-10). Tracking issue: [#88](https://github.com/zjshen14/opencli/issues/88). Phase: [Roadmap A4](../roadmap.md)._
+_Status: Implemented — merged in 3ee9367 (2026-05-10). Tracking issue: [#88](https://github.com/zjshen14/opencli/issues/88). Phase: [Roadmap A4](../roadmap.md)._
 
 ---
 
--- a/docs/design/a4-file-mentions-history.md
+++ b/docs/design/a4-file-mentions-history.md
@@ -1,6 +1,6 @@
 # Design: A4 — @file mention expansion + per-CWD input history
 
-_Status: Ready for implementation. Tracking issue: [#88](https://github.com/zjshen14/opencli/issues/88). Phase: [Roadmap A4](../roadmap.md)._
+_Status: Implemented — merged in <sha> (2026-05-10). Tracking issue: [#88](https://github.com/zjshen14/opencli/issues/88). Phase: [Roadmap A4](../roadmap.md)._
 
 ---
 
--- a/src/cli/input.test.ts
+++ b/src/cli/input.test.ts
@@ -1,10 +1,26 @@
-import { describe, it, expect } from "vitest";
-import {
+import { describe, it, expect, afterEach, vi } from "vitest";
+import { rm } from "node:fs/promises";
+import { join } from "node:path";
+import os from "node:os";
+
+// Isolate history writes from the real ~/.opencli
+const tmpHome = join(os.tmpdir(), `opencli-input-test-${Date.now()}`);
+vi.mock("node:os", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:os")>();
+  return { ...actual, default: { ...actual, homedir: () => tmpHome }, homedir: () => tmpHome };
+});
+
+// Import after mock so AGENT_DIR resolves to tmpHome.
+// All input.ts exports use the dynamic import to avoid static-import hoisting
+// (which would trigger homedir() before tmpHome is initialized).
+const {
+  loadHistory,
+  saveHistory,
   insertAtCursor,
   deleteBeforeCursor,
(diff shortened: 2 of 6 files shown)


zjshen14 (commented):
## Review — A4 @file mention expansion + per-CWD history

**Verdict: Approve with one fix needed.**

---

### What I checked

**`src/cli/mentions.ts`**

- Fast-path `input.includes("@")` check avoids the regex entirely — good. ✓
- Reverse-order replacement application correctly preserves earlier offsets when multiple tokens exist in one input. ✓
- Binary detection (`\0` byte in first 8 000 bytes) matches the spec. ✓
- `expandFile`: `ENOENT` special-cased to a friendlier message; other errors pass through. ✓
- `expandGlob`: file-count cap and total-char cap both correctly set `capped = true`; double-warning on chars cap with 0 blocks is unreachable (per-file 50K cap makes any block < 200K total). ✓
- `walk` / `matchGlob` correctly duplicated from `glob.ts` — no coupling to tool internals. ✓
- `{` in glob detection triggers glob mode, but `matchGlob` escapes `{` so brace expansion never matches. Consistent limitation with `glob.ts`; worth a README note eventually but not a blocker.

**`src/cli/input.ts`**

`historyFile(cwd)` → `~/.opencli/history/<base64url>`. Signatures updated, `mkdir` now creates the `history/` subdirectory. ✓

**`src/cli/input.test.ts`**

`vi.mock("node:os")` + dynamic `await import("./input.js")` correctly ensures `AGENT_DIR` resolves to the temp dir before module-level code runs. Per-CWD isolation test, round-trip, missing-file, and 500-cap tests all present. ✓

**`src/cli/mentions.test.ts`**

All 9 design-specified test cases present and using real temp fs. ✓

---

### Bug — session log saves unexpanded input

`repl.ts` line 68:
```typescript
void session.log({ type: "user", content: rawInput });  // ← rawInput, not expanded
```

`session.log` is called before the expansion runs (line 73: `const input = expanded`). The agent receives the expanded content, but the session JSONL gets the raw `@token`. When a user resumes the session with `--resume`, the reconstructed conversation history will contain the literal `@src/core/agent.ts` token rather than the file contents — the agent won't have the context it had in the original session.

**Fix:** move the `session.log` call to after line 73 and log `input` (expanded):

```typescript
// Expand @file/@glob mentions before passing to agent or slash commands
const { expanded, warnings } = await expandMentions(rawInput, cwd);
for (const w of warnings) printInfo(w);
const input = expanded;

// Log expanded content so resumed sessions have the same context
void session.log({ type: "user", content: input });
```

`history.unshift(rawInput)` stays — compact history is correct. Only the session log needs the expanded form.

---

### Minor: capped-glob warning message

When the file-count cap is hit, the warning is:
```
@src/**/*.ts: expansion capped at 20 files
```
(`blocks.length` equals `MAX_GLOB_FILES` at the point of the warning.) ✓ Matches the spec.

zjshen14 (commented):
Fixed in afec31c — moved `session.log` to after expansion so the session JSONL records the fully expanded content. `history.unshift(rawInput)` stays compact as before.

PR #97, opened 2026-05-10
feat(core): context compaction via LLM summarization (A5)

### LINKED ISSUE #26: Feature: context compaction via LLM summarization instead of hard-dropping history
Part of #19 (ReAct loop quality improvements).

## What

When conversation history approaches the context window limit, summarize older turns using the LLM instead of hard-dropping them. This allows very long sessions to continue coherently without losing critical decisions.

## Current behaviour

`ContextManager.prune()` uses `slice(-maxHistoryMessages)` — old messages are dropped entirely. On a 50-message default, a long session loses its original user request and early decisions, causing the agent to drift or repeat work.

## Proposal

Before dropping messages, call the LLM with a compact summarization prompt:

```typescript
async function compactHistory(
  messages: Message[],
  client: LLMClient,
): Promise<Message> {
  const summary = await client.complete(
    "Summarize the key decisions, files changed, errors encountered, and current state from this conversation. Be concise — under 200 words. Preserve file paths and function names exactly.",
    messages,
  );
  return {
    role: "user",
    parts: [{ type: "text", text: `[Session summary — earlier context compacted]\n${summary}` }],
  };
}
```

Replace the dropped tail with this single synthetic message. The agent retains the gist of earlier work without the full token cost.

## Implementation notes

- Trigger at 80% of `maxHistoryMessages` (soft threshold), not at 100% (hard cut)
- Use a cheap/fast model for summarization (e.g., flash-lite) rather than the active session model
- Gate behind `--compaction` flag initially for testing; promote to default once validated
- Add a `PreCompact` hook (like Claude Code's) that fires before summarization, allowing users to archive the full transcript

## Reference

- Claude Code: five-layer budget reduction → auto-compact pipeline; `PreCompact` hook for archiving
- Cline: "Auto Compact" at 80% context utilisation, middle-removal with summarization
- OpenHands: `LLMSummarizingCondenser` — reported 2× cost reduction with no benchmark degradation

zjshen14:
## Summary

Closes #26.

Implements Phase A5 — context compaction via LLM summarization, replacing the hard-drop `slice(-N)` behaviour in `ContextManager.prune()` with a graceful summarization pass when history approaches the limit.

- **`ContextManager`** gains three new methods: `needsCompaction()` (returns true at 80% of `maxHistoryMessages`), `getCompactableMessages()` (the older head to summarize), and `replaceWithSummary(msg)` (swaps the head for a single summary message, keeping the 40% tail verbatim). The existing hard-cap `prune()` remains as a safety net.
- **`Agent`** checks `needsCompaction()` at the top of each iteration loop and calls a private `compact()` method that streams from the LLM with a dedicated summarization system prompt (no tools, no full system instruction). The resulting text is wrapped in a `[Session summary — earlier context compacted]` user message.
- **CLI flag**: `opencli chat --compaction` enables the feature. Off by default so existing behaviour is unchanged.

## Test plan

- [x] `ContextManager.needsCompaction()` returns false below 80% threshold and true at/above it
- [x] `ContextManager.getCompactableMessages()` returns the head (history minus 40% tail)
- [x] `ContextManager.replaceWithSummary()` collapses to summary + tail; subsequent `needsCompaction()` returns false
- [x] Agent calls the LLM summarizer when threshold is reached (verified with mock client)
- [x] Agent does NOT call summarizer when `compactionEnabled` is false (default)
- [x] All 464 existing tests continue to pass

https://claude.ai/code/session_01XMkecDE5XsrPgdosUzsDFR

---
_Generated by [Claude Code](https://claude.ai/code/session_01XMkecDE5XsrPgdosUzsDFR)_

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -40,6 +40,10 @@ program
   .option("--sandbox <mode>", "Sandbox mode for bash tool: auto | strict | off (default: auto)")
   .option("--provider <provider>", "Override provider detection: gemini | anthropic | openai")
   .option("--base-url <url>", "Custom base URL for proxy or local inference (e.g. LiteLLM)")
+  .option(
+    "--compaction",
+    "Enable context compaction: summarize old messages via LLM instead of dropping them",
+  )
   .action(async (opts) => {
     const sessionId = opts.session ?? (opts.resume ? "latest" : undefined);
     await startChat(
@@ -50,6 +54,7 @@ program
       opts.sandbox as string | undefined,
       opts.provider as string | undefined,
       opts.baseUrl as string | undefined,
+      opts.compaction as boolean | undefined,
     );
   });
 
@@ -167,6 +172,7 @@ async function startChat(
   sandboxFlag?: string,
   providerOverride?: string,
   baseUrlOverride?: string,
+  compaction?: boolean,
 ): Promise<void> {
   const { agent, skills, mcpManager, snapshotManager } = await createAgent(
     modelOverride,
@@ -175,6 +181,7 @@ async function startChat(
     sandboxFlag,
     providerOverride,
     baseUrlOverride,
+    compaction,
   );
 
   const cleanup = async () => {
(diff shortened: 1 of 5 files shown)


zjshen14:
Superseded by PR #114 (A5a context compaction with /compact and /context commands). The v1 implementation landed at commit 4aba15e with cleaner design and better test coverage.

PR #122, opened 2026-05-17
fix(prune): prevent model-first history when fallback fires

### LINKED ISSUE #120: Bug: prune() fallback returns model-first history, causing Gemini INVALID_ARGUMENT crash
## What happened

When a session runs long enough that the entire `slice(-maxHistoryMessages)` window is orphaned (all model turns or user-with-results turns, no clean user text message), `prune()` falls back to returning the raw slice rather than an empty array. If that slice starts with a model-role message containing function calls, Gemini rejects the next API call with:

> Please ensure that function call turn comes immediately after a user turn or after a function response turn.

This was observed in a real session where the agent kept re-running a failing test, the history grew past 50 messages, and the prune fallback produced a model-first window.

## Root cause

In \`src/core/context.ts\`, the fallback in \`prune()\`:

\`\`\`typescript
this.history = startIdx < sliced.length ? sliced.slice(startIdx) : sliced;
\`\`\`

When \`startIdx >= sliced.length\` (entire window orphaned), it returns \`sliced\` unmodified. If \`sliced[0].role === "model"\`, Gemini rejects the next request.

## Fix

When the full-orphan fallback fires, scan from the **tail** of the slice for the last clean user message and use that as the start point. If no clean user message exists anywhere in the slice, keep only the final message as a minimal state rather than returning an invalid model-first window.

## Impact

Hard crash — the session terminates with an unhandled provider error after context grows past \`maxHistoryMessages\`.

zjshen14:
## Summary

- Fixes a hard crash observed in production: when a single user turn triggers so many tool-call/result pairs that the original user message scrolls out of the `slice(-maxHistoryMessages)` window, the previous fallback returned the raw slice — which could start with a `role: "model"` message that Gemini rejects with `INVALID_ARGUMENT: Please ensure that function call turn comes immediately after a user turn or after a function response turn.`
- New behavior: when the head scan finds no clean start, do a **tail scan** for the last clean user text message and use that as the window start. If even that fails (window is entirely tool-call/result pairs with no user text at all — pathological with very small `maxHistoryMessages`), prepend a synthetic user anchor so history never starts with a model turn.
- 2 new tests covering both paths.

Closes #120

## Test plan

- [ ] `npm run typecheck && npm run lint && npm run format:check && npm test` — 503 tests pass
- [ ] Simulate crash: set `historySize: 4` in config, run a session where the agent makes 5+ consecutive tool calls in response to one user message — should no longer crash with INVALID_ARGUMENT; if context is truncated, the session continues with a tail-anchored or synthetic-anchored window

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,6 +7,10 @@ It must be kept in sync with `CLAUDE.md` — see `docs/engineering-practices.md`
 
 OpenCLI — an open-source AI agent CLI that supports Google Gemini and Anthropic Claude models, modeled after Claude Code. The implementation is a working prototype in TypeScript/Node.js (ESM, Node 20+).
 
+## Getting Started with a Task
+
+**New to this project?** Start here: [Task Workflow](docs/task-workflow.md) — it walks you through the typical task lifecycle, from reading the issue to committing your work. It references this document and [docs/engineering-practices.md](docs/engineering-practices.md) for the conventions you need to know.
+
 ## Commands
 
 ```bash
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,6 +8,10 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 OpenCLI — an open-source AI agent CLI that supports Google Gemini and Anthropic Claude models, modeled after Claude Code. The implementation is a working prototype in TypeScript/Node.js (ESM, Node 20+).
 
+## Getting Started with a Task
+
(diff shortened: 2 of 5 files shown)


zjshen14:
Changes already landed on main (commits 7ad3e22, c8125c5). Closing stale PR.

zjshen14:
Reopening — the prune fix wasn't actually landed on origin/main (it was only on local). The branch is now rebased on top of the latest dependabot updates and includes the task-workflow doc as well. All 503 tests pass.

PR #124, opened 2026-05-17
fix(agent): abort loop on repeated environmental errors

### LINKED ISSUE #121: Bug: agent loops on environmental errors (EPERM, EACCES) instead of diagnosing and stopping
## What happened

When a tool result contains an OS-level error that code changes cannot fix (e.g. \`EPERM: operation not permitted\`, \`EACCES\`), the agent keeps making different edits and re-running the failing command across many turns. The stuck-loop detector only catches *identical* consecutive call signatures — varied-but-futile edit attempts slip through.

Real example: \`listen EPERM: operation not permitted 0.0.0.0\` is a macOS network binding restriction. The agent edited \`src/index.ts\` five times across multiple turns, each time slightly differently, then re-ran \`npm test\` — never diagnosing the environmental root cause.

## Root cause

\`STUCK_THRESHOLD\` in \`src/core/agent.ts\` only fires when the *same* tool + args appear 3 times in a row. Different edits with the same underlying error bypass it entirely.

## Proposed fix

Add a second signal in the agentic loop: if the same error substring (matched against a denylist: \`EPERM\`, \`EACCES\`, \`permission denied\`, \`operation not permitted\`) appears in consecutive \`tool_result\` outputs for N turns (suggested: 3), yield an \`error\` event with a diagnosis message and stop the loop.

This avoids burning turns and API cost on changes that cannot fix environmental restrictions.

## Impact

Poor UX — session runs until context limit is hit and crashes, instead of surfacing a clear diagnosis early.

zjshen14:
## Summary

- Adds a second stuck-loop signal to the agentic loop for OS-level errors that code changes cannot fix. If the same error pattern (`EPERM`, `EACCES`, `permission denied`, `operation not permitted`, `access is denied`) appears in tool results across 3 consecutive turns, the loop stops and yields a clear diagnostic error message.
- The existing stuck-loop guard catches *identical* call signatures — this new guard catches varied-but-futile edit attempts where the agent modifies different files each turn but keeps hitting the same OS restriction.
- Counter resets when a turn produces no matching error, so brief transient failures don't trigger it.
- Adds `env_error_loop` to the `guard_triggered` observability event union.

**Real trigger:** `listen EPERM: operation not permitted 0.0.0.0` — a macOS network binding restriction that caused 5+ edit/test cycles before the session crashed.

Closes #121

## Test plan

- [ ] `npm run typecheck && npm run lint && npm run format:check && npm test` — 507 tests pass
- [ ] EPERM aborts after 3 turns with diagnostic message
- [ ] EACCES aborts after 3 turns
- [ ] Counter resets on a clean turn — no false abort
- [ ] Single isolated env error → loop completes normally, no abort

### DIFF
--- a/src/core/agent.test.ts
+++ b/src/core/agent.test.ts
@@ -169,6 +169,137 @@ describe("Agent plan mode", () => {
   });
 });
 
+describe("Agent environmental error guard", () => {
+  // Client that makes a different call each turn (varying args) so stuck-loop
+  // detection never fires — mirrors the real scenario where the agent edits
+  // different files each turn but the underlying OS error persists.
+  function makeVaryingClient(toolName: string): { client: LLMClient; getCallCount: () => number } {
+    let n = 0;
+    const client: LLMClient = {
+      async *stream() {
+        n++;
+        yield { type: "function_call", id: `c${n}`, name: toolName, args: { n } } as StreamEvent;
+        yield { type: "done" } as StreamEvent;
+      },
+    };
+    return { client, getCallCount: () => n };
+  }
+
+  it("aborts after 3 consecutive turns with the same EPERM error", async () => {
+    const { client } = makeVaryingClient("bash");
+    const registry = new ToolRegistry();
+    registry.register({
+      name: "bash",
+      description: "",
+      parameters: { type: "object", properties: { n: { type: "number" } } },
+      execute: async () => ({
+        success: false,
+        output: "",
+        error: "Error: listen EPERM: operation not permitted 0.0.0.0",
+      }),
+    });
+
+    const agent = new Agent(client, registry, new SkillRegistry(), undefined, undefined, 100);
+    const events = await collectEvents(agent, "run tests");
+    const error = events.find((e) => e.type === "error");
+    expect(error).toBeDefined();
+    expect((error as { type: "error"; message: string }).message).toMatch(/EPERM/i);
+    expect((error as { type: "error"; message: string }).message).toMatch(/environment/i);
+    // Should stop after exactly 3 turns
+    const toolCalls = events.filter((e) => e.type === "tool_call");
+    expect(toolCalls).toHaveLength(3);
+  });
+
+  it("aborts on EACCES pattern", async () => {
+    const { client } = makeVaryingClient("bash");
+    const registry = new ToolRegistry();
+    registry.register({
+      name: "bash",
+      description: "",
+      parameters: { type: "object", properties: { n: { type: "number" } } },
+      execute: async () => ({
+        success: false,
+        output: "",
+        error: "Error: EACCES: permission denied, open '/etc/shadow'",
+      }),
+    });
+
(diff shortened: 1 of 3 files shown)


zjshen14:
## Review: ✅ Approve

**Strengths:**

1. **Real bug, focused fix.** The motivation (`listen EPERM` burning 5+ edit/test cycles) is a concrete pain point worth fixing. The 3-turn threshold + counter-reset is a sound design — catches genuine env-error loops without false-triggering on transient failures.

2. **Mirrors the existing stuck-loop pattern.** `envErrorPattern` / `envErrorCount` parallels `lastCallSig` / `stuckCount` exactly, which makes the code immediately readable for anyone familiar with the existing guard.

3. **Excellent test coverage.** All four critical cases are covered:
   - 3 consecutive EPERM → aborts ✅
   - EACCES variant matches ✅
   - Clean turn resets counter (no false abort) ✅
   - Single isolated error doesn't trigger ✅
   - The `makeVaryingClient` trick (varying args to bypass stuck-loop) is a clean way to isolate the new guard.

4. **Observability union update is proper.** The `guard_triggered` event correctly extends to include `env_error_loop`.

5. **Diagnostic message is actionable.** "check permissions, network settings, or sandbox configuration" gives the user a concrete path forward instead of a generic abort.

**Minor follow-up suggestions (non-blocking):**

- **Pattern list could include resource-exhaustion errors** like `ENOSPC` (disk full), `EMFILE` (too many open files), `ENOMEM` (out of memory). Same logic applies — code changes can't fix them. Worth a follow-up issue if it becomes a real problem.
- **Edge case** — if the error message format varies across turns (e.g., turn 1 has `EPERM` only, turn 2 has `operation not permitted` only), the counter resets because the matched pattern differs. In practice, error messages are usually consistent, but worth noting.
- **No test for the observability event firing** — current tests verify behavior (event emitted), but could add a direct check on the `guard_triggered` event with `guard: "env_error_loop"`.

These are all nice-to-haves, not blockers. The PR is tight, well-scoped, and merges cleanly. CI is green, all 507 tests pass.

Ready to merge.

zjshen14:
Addressed the three review suggestions:
- Added `ENOSPC`, `EMFILE`, `ENOMEM` and their text variants (`no space left on device`, `too many open files`) to the pattern list
- Added a test verifying the `guard_triggered{env_error_loop}` observability event fires
- The pattern-mismatch edge case (different error strings across turns) is noted in the issue as a known limitation — acceptable given how consistent real error messages are in practice
508 tests passing.

zjshen14:
✅ **Approved and merging.**

Both of my main suggestions addressed in commit 8dd0bfa:
- Pattern list expanded with resource-exhaustion errors (`ENOSPC`, `EMFILE`, `ENOMEM` + text variants)
- New test verifies the `env_error_loop` `guard_triggered` observability event fires

The remaining edge case (varying error format across turns) is a non-blocking nit — file a follow-up issue if it ever bites in practice.

PR #130, opened 2026-05-18
validate --max-turns and --temperature flags to prevent silent NaN

### LINKED ISSUE #116: Numeric CLI flags accept garbage values without validation
## Summary

\`--max-turns\` and \`--temperature\` are parsed with bare \`parseInt\` / \`parseFloat\`. Non-numeric values produce silent \`NaN\` and are passed through to the agent / provider. Out-of-range values are also accepted.

## Reproductions

\`\`\`bash
# 1. Non-numeric max-turns → NaN; the safety guard 'turns > maxTurns' is always false
opencli run \"hi\" --yes --max-turns abc

# 2. Non-numeric temperature → NaN passed to provider
opencli run \"hi\" --yes --temperature notanumber

# 3. Out-of-range temperature
opencli run \"hi\" --yes --temperature -1
opencli run \"hi\" --yes --temperature 999
\`\`\`

All four execute without complaint (assuming a working API key).

## Why this matters

- **\`--max-turns NaN\`** disables the safety guard in [src/core/agent.ts:159](src/core/agent.ts#L159) (\`if (turns > this.maxTurns)\` is always false when \`maxTurns\` is \`NaN\`). The stuck-loop detector still fires, but the cost-bounding guard is silently neutered.
- **\`--temperature NaN\`** is sent to the provider; downstream behaviour is provider-dependent and may produce a confusing error far from the cause.
- **Out-of-range temperature** depends on provider tolerance — some accept \`>1\`, some don't.

## Root cause

[src/cli/index.ts:38, 76, 83](src/cli/index.ts#L38) use \`parseInt\` / \`parseFloat\` directly as the commander coercer. Neither library validates the result.

## Suggested fix

Wrap each in a validator:

\`\`\`ts
function parseTurns(raw: string): number {
  const n = parseInt(raw, 10);
  if (!Number.isFinite(n) || n < 1) throw new InvalidArgumentError('--max-turns must be a positive integer');
  return n;
}

function parseTemperature(raw: string): number {
  const n = parseFloat(raw);
  if (!Number.isFinite(n) || n < 0 || n > 2) throw new InvalidArgumentError('--temperature must be between 0 and 2');
  return n;
}
\`\`\`

Commander already does clean error formatting for \`InvalidArgumentError\`. Same pattern as the existing \`--provider\` and \`--sandbox\` validation, which both work correctly.

Found while dogfooding the CLI.

zjshen14:
## Summary

- Replaces bare `parseInt`/`parseFloat` coercers on `--max-turns` and `--temperature` with validator functions that throw `InvalidArgumentError` on bad input
- `--max-turns abc` or `--max-turns 0` now exits 1 with `"--max-turns must be a positive integer"` instead of silently passing `NaN` (which disabled the safety guard in `agent.ts`)
- `--temperature notanumber` or out-of-range values now exits 1 with `"--temperature must be between 0 and 2"` instead of forwarding `NaN` to the provider
- Pattern matches existing `--sandbox` / `--provider` validation already in the codebase

Closes #116

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — all pass
- [x] New smoke tests cover: non-numeric `--max-turns`, zero `--max-turns`, non-numeric `--temperature`, out-of-range `--temperature` (all run against the built dist via `npm run build`)
- [x] Valid values (`--max-turns 5`, `--temperature 0.7`) still pass through unchanged

---
_Generated by [Claude Code](https://claude.ai/code/session_01BLp3cPtmy4GeAVwDTSqipY)_

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -1,4 +1,4 @@
-import { Command } from "commander";
+import { Command, InvalidArgumentError } from "commander";
 import { createClient, createCompactionClient, detectProvider } from "../providers/factory.js";
 import { Agent } from "../core/agent.js";
 import { createDefaultRegistry } from "../tools/index.js";
@@ -19,6 +19,20 @@ import { McpManager } from "../mcp/manager.js";
 import { registerMcpCommand } from "./mcp-cmd.js";
 import { SnapshotManager } from "../state/snapshot.js";
 
+function parseTurns(raw: string): number {
+  const n = parseInt(raw, 10);
+  if (!Number.isFinite(n) || n < 1)
+    throw new InvalidArgumentError("--max-turns must be a positive integer");
+  return n;
+}
+
+function parseTemperature(raw: string): number {
+  const n = parseFloat(raw);
+  if (!Number.isFinite(n) || n < 0 || n > 2)
+    throw new InvalidArgumentError("--temperature must be between 0 and 2");
+  return n;
+}
+
 const program = new Command();
 
 program
@@ -35,7 +49,7 @@ program
   )
   .option("-r, --resume", "Resume the most recent session")
   .option("-s, --session <id>", "Resume a specific session by ID")
-  .option("--max-turns <n>", "Maximum agent iterations per prompt (default: 50)", parseInt)
+  .option("--max-turns <n>", "Maximum agent iterations per prompt (default: 50)", parseTurns)
   .option("--debug", "Emit structured observability events to stderr as JSON")
   .option("--sandbox <mode>", "Sandbox mode for bash tool: auto | strict | off (default: auto)")
   .option("--provider <provider>", "Override provider detection: gemini | anthropic | openai")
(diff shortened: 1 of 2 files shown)


zjshen14:
Good fix — `InvalidArgumentError` gives commander the clean exit-1 formatting that `parseInt`/`parseFloat` can't, and the rejection of `0` for `--max-turns` is the right call.

Two things to check before landing:

**Coverage gap: `chat` command `--temperature`**
The diff updates `--temperature` on the `run` subcommand but I don't see a corresponding change on the `chat` command definition. If `chat` also accepts `--temperature` (and the current code uses bare `parseFloat` there), the fix is incomplete — a user could still pass `--temperature abc` to `chat` and get a silent `NaN`. Worth confirming whether `chat` has that option and, if so, applying `parseTemperature` there too.

**Stray comma in `src/cli/run.smoke.test.ts`**
After the third added test case (`"run --temperature with non-numeric value exits 1..."`), the closing line reads:
```ts
  }),
```
with a trailing comma instead of a semicolon. TypeScript parses this as a comma expression (the return value of `it()` is discarded, so Vitest still registers the test), but it reads like a mistake and a linter configured with `no-comma-dangle` in statement position would flag it.

Recommendation: confirm `chat --temperature` coverage, fix the comma, then merge.



---

PR #148, opened 2026-05-23
fix(sandbox): loosen auto profile so normal coding workflows succeed

### LINKED ISSUE #127: Sandbox auto mode is too restrictive for real coding workflows (npm/pip/git bootstrap)

zjshen14:
## Summary

- Loosens the default `auto` sandbox profile so `npm install`, `pip install`, `gh`, `git clone`, and `curl` all succeed without forcing users to `--sandbox off`.
- Adds writes for XDG dirs + common package-manager dot-dirs (`~/.npm`, `~/.cache`, `~/.cargo`, etc.), allows all outbound network, allows writes to `/dev` (so `curl -o /dev/null` and shell redirects work), and adds `(allow process-info* (target others))` for non-setuid process tools.
- Updates the docs to be honest about what `auto` buys: prevent accidental writes to system & credential paths. **Not a security boundary** — `strict` (still stubbed; Phase 2 in [docs/design/a7-sandbox-loosen-auto.md](docs/design/a7-sandbox-loosen-auto.md)) is the real isolation mode.
- Linux: drops `--unshare-net`; pre-creates the dot-dirs at runner construction so `--bind` doesn't fail on first run.

**Still blocked:** writes to `~/.ssh`, `~/.aws`, `~/.gnupg`, `~/.docker`, `~/.kube`, `~/Documents`, `~/Desktop`, `/etc`, `/usr`, `/System`, `/Applications`. The deny-default catches everything outside the explicit allow list.

**Note on `/bin/ps`:** the original issue flagged `ps` as blocked. It's setuid root — macOS sandbox-exec refuses to exec setuid binaries regardless of profile. Documented as a platform limitation; non-setuid tools (`pgrep`, `lsof -p`) work fine.

## Related issue

Closes #127

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — 534 tests pass
- [x] Manual end-to-end: `npm install lodash` from a fresh dir succeeds inside the sandbox (fetches from registry.npmjs.org, writes to `~/.npm/_cacache`, creates `node_modules/lodash/`)
- [x] Existing sandbox tests still pass: writes to `/etc/hosts` denied, writes to CWD allowed, loopback bind/connect works
- [x] New tests cover the new allows: writes to `~/.npm`, `~/.cache`, `~/Library/Caches` succeed; writes to `~/.ssh`, `~/.aws` denied; `curl https://example.com` succeeds; `pgrep` works
- [ ] Linux CI passes (bwrap changes — verified via review since macOS dev machine)

## Notes for reviewers

The behavior change is significant — outbound network is now allowed by default. Anyone relying on `auto` to block external network needs to know:
1. `auto` was never really a security boundary (file-read* exposed everything)
2. `strict` mode (Phase 2, separate PR) will provide real isolation
3. In the interim, `--sandbox off` + external firewall is the recommended setup for security-sensitive contexts

The design doc ([docs/design/a7-sandbox-loosen-auto.md](docs/design/a7-sandbox-loosen-auto.md)) lays out the full reasoning, Phase 2 strict mode design, and decision matrix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,8 +190,10 @@ The `bash` tool runs inside an OS-level sandbox by default (`--sandbox auto`):
 | Mode | Behaviour |
 |------|-----------|
 | `auto` (default) | Prevents accidental writes to system & credential paths (`/etc`, `~/.ssh`, `~/.aws`, etc.). Reads and network unrestricted. Writes allowed inside CWD, `/tmp`, and common dev dirs (`~/.npm`, `~/.cache`, `~/.cargo`, `~/Library/Caches`, …). **Not a security boundary** — use `strict` for real isolation. |
-| `strict` | Real isolation: no external network, writes only to CWD + tmp, reads restricted to CWD + system binaries. **Currently stubbed — falls back to `auto` with a warning.** Tracked in [#127](https://github.com/zjshen14/opencli/issues/127). |
+| `strict` | Real isolation: no external network, writes only to CWD + tmp, reads restricted to CWD + system binaries. **Currently stubbed — falls back to `auto` with a warning.** Tracked in [#149](https://github.com/zjshen14/opencli/issues/149). |
 | `off` | No sandbox |
+
+> **⚠ Behavior change (May 2026):** Prior to this release, `--sandbox auto` denied all external network access. As of [#127](https://github.com/zjshen14/opencli/issues/127), `auto` allows external network by default — every real coding workflow (`npm install`, `gh`, `git clone`, `curl`) was blocked otherwise. If you relied on the previous network-deny behavior, use `--sandbox strict` (once [#149](https://github.com/zjshen14/opencli/issues/149) lands) or `--sandbox off` plus an external firewall.
 
 ```bash
 # CLI flag
--- a/README.md
+++ b/README.md
@@ -183,14 +183,14 @@ Environment variables take precedence over config file:
 
 The `bash` tool runs inside an OS-level sandbox by default (`--sandbox auto`):
 
-- **macOS** — uses `sandbox-exec` (built-in, no install required). Network access is denied; writes outside the project root and `/tmp` are denied.
-- **Linux** — uses `bwrap` (bubblewrap) via user namespaces. Same isolation contract. Falls back to passthrough with a warning if `bwrap` is not installed.
+- **macOS** — uses `sandbox-exec` (built-in, no install required). Writes outside common dev locations are denied; reads and network are unrestricted.
+- **Linux** — uses `bwrap` (bubblewrap) via user namespaces. Same contract. Falls back to passthrough with a warning if `bwrap` is not installed.
 - **Windows / other** — no native sandbox; runs without isolation with a warning.
 
 | Mode | Behaviour |
 |------|-----------|
-| `auto` (default) | Network denied; writes allowed only inside CWD and `/tmp` |
-| `strict` | Falls back to `auto` with a warning (full isolation planned) |
+| `auto` (default) | Prevents accidental writes to system & credential paths (`/etc`, `~/.ssh`, `~/.aws`, etc.). Reads and network unrestricted. Writes allowed inside CWD, `/tmp`, and common dev dirs (`~/.npm`, `~/.cache`, `~/.cargo`, `~/Library/Caches`, …). **Not a security boundary** — use `strict` for real isolation. |
+| `strict` | Real isolation: no external network, writes only to CWD + tmp, reads restricted to CWD + system binaries. **Currently stubbed — falls back to `auto` with a warning.** Tracked in [#127](https://github.com/zjshen14/opencli/issues/127). |
 | `off` | No sandbox |
 
 ```bash
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -324,6 +324,14 @@ interface SandboxRunner {
 | Linux | `BwrapRunner` (bwrap) | `PassthroughRunner` |
 | Other | `PassthroughRunner` + warning | `PassthroughRunner` |
 
+**Mode semantics** (see [docs/design/a7-sandbox-loosen-auto.md](design/a7-sandbox-loosen-auto.md)):
+
+| Mode | Purpose | Trust model |
+|------|---------|-------------|
+| `auto` | Prevent obvious accidents — writes to system & credential paths denied; reads and network unrestricted; writes allowed to CWD, `/tmp`, and common dev dot-dirs (`~/.npm`, `~/.cache`, etc.). | Convenience first. **Not a security boundary.** |
+| `strict` | Real isolation — no external network, writes only to CWD + tmp, reads restricted to CWD + system binaries. Currently stubbed; falls back to `auto`. | Untrusted code; expect friction. |
+| `off` | No isolation. | Trusted local environment. |
+
 **Startup warning**: `runner.warning` is emitted exactly once to `stderr` at CLI startup if isolation was downgraded (e.g. bwrap missing, namespaces disabled, unsupported platform).
 
 #### Tool interface
--- a/docs/design/a7-sandbox-loosen-auto.md
+++ b/docs/design/a7-sandbox-loosen-auto.md
@@ -67,11 +67,10 @@ File: [src/tools/exec/sandbox/sandbox-exec.ts](../../src/tools/exec/sandbox/sand
 (allow process*)
 (allow signal)
 (allow sysctl-read)
-(allow sysctl-write)
 (allow mach*)
 (allow ipc*)
 
-; NEW: process introspection (fixes /bin/ps block)
+; NEW: process introspection (enables non-setuid tools like pgrep)
 (allow process-info* (target others))
 
 ; Reads: allow everywhere (unchanged)
@@ -82,9 +81,16 @@ File: [src/tools/exec/sandbox/sandbox-exec.ts](../../src/tools/exec/sandbox/sand
 (allow file-write* (subpath "/private/tmp"))
 (allow file-write* (subpath "/private/var/folders"))
 
-; NEW: standard device nodes (/dev/null, /dev/stdout, /dev/tty, etc.)
-; Required for `curl -o /dev/null`, shell redirects, many other tools.
-(allow file-write* (subpath "/dev"))
+; NEW: standard device nodes (explicit literals to keep allow surface narrow)
+; Required for curl -o /dev/null, shell redirects, /dev/tty prompts.
+(allow file-write*
+  (literal "/dev/null")
+  (literal "/dev/zero")
+  (literal "/dev/stdout")
+  (literal "/dev/stderr")
+  (literal "/dev/tty")
+  (literal "/dev/urandom")
+  (literal "/dev/random"))
 
 ; NEW: XDG base directories
 (allow file-write* (subpath "${HOME}/.cache"))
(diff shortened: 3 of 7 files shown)


zjshen14:
## Review

Read the full diff (588 additions, 7 files) including the 410-line design doc. Strong execution overall — going through it the same way I've reviewed earlier designs.

## Strengths

**Faithful to the issue.** #127 proposed Option A (loosen auto, reserve strict for real isolation). This PR implements exactly that and splits Phase 2 (strict) cleanly into a follow-up — good engineering. Phase 1 ships immediately useful unblock; Phase 2 gets its own review surface.

**Design doc is the right kind of thorough.** The mode-taxonomy table, failure-modes matrix, and \"Decisions to confirm\" section are exemplary. The honest framing — \"auto is theatre, here's why\" — is rare and welcome. The industry comparison (Cursor / Cline / Claude Code) lands the right precedent.

**Conservative dot-dir list.** Allows the universal package managers (\`~/.npm\`, \`~/.cargo\`, \`~/.gem\`, \`~/.gradle\`, \`~/.m2\`, \`~/.rustup\`, \`~/.yarn\`, \`~/.pnpm-store\`) plus XDG (\`~/.cache\`, \`~/.config\`, \`~/.local\`) plus macOS \`~/Library/Caches\`. Explicitly does NOT allow \`~/.ssh\`, \`~/.aws\`, \`~/.gnupg\`, \`~/.docker\`, \`~/.kube\`. Tests cover both directions — good restraint.

**Linux pre-create logic is correct.** \`bwrap --bind\` fails if the source path is absent, so the synchronous \`mkdirSync\` in the constructor solves a real first-run issue. The error-swallow is appropriate.

**Honest about \`/bin/ps\`.** The PR acknowledges that \`/bin/ps\` and \`/usr/bin/top\` are setuid root and macOS sandbox-exec refuses setuid execs regardless of profile. The pgrep test verifies process introspection works in principle while documenting the platform limitation. Better to be honest than to claim a fix that doesn't materialize.

## Concerns

### 1. \`sysctl-write\` allowance is unjustified

\`\`\`scheme
(allow sysctl-write)
\`\`\`

The A1 profile didn't have this. The design doc lists it under \"Process lifecycle\" but doesn't explain *why* it's needed. \`sysctl-write\` lets a process modify kernel parameters — not something a typical \`npm install\` / \`pip install\` does. None of the design doc's enumerated workflows (npm/pip/cargo/gh/git/curl) need it.

Two possibilities:
- Added defensively without a specific use case → remove it
- Added because *something* during testing failed without it → name what, and confirm the allow is the right granularity (could be a more specific \`(allow sysctl-write (sysctl-name \"...\"))\`)

Either way, the PR should justify or drop this line. Adding capabilities silently to a profile change like this is exactly the kind of drift that erodes the \"auto isn't a security boundary, but it prevents obvious mistakes\" framing.

### 2. \`(allow file-write* (subpath \"/dev\"))\` is broader than the comment implies

The comment says \"/dev/null, /dev/stdout, /dev/tty, etc.\" But \`subpath \"/dev\"\` covers:
- \`/dev/disk*\`, \`/dev/rdisk*\` — block device nodes
- \`/dev/audio*\`, \`/dev/video*\` — capture devices on some platforms
- \`/dev/console\`, \`/dev/kmsg\` — kernel-facing interfaces

In practice these are protected by user-level permissions (you can't write to \`/dev/disk0\` without root), so the practical security impact is nil. But the comment understates what the rule actually allows.

The narrow version would be more honest:

\`\`\`scheme
(allow file-write*
  (literal \"/dev/null\")
  (literal \"/dev/zero\")
  (literal \"/dev/stdout\")
  (literal \"/dev/stderr\")
  (literal \"/dev/tty\")
  (literal \"/dev/urandom\")
  (literal \"/dev/random\"))
\`\`\`

This covers every legitimate use the comment cites (\`curl -o /dev/null\`, shell redirects to stdout/stderr, /dev/tty for prompts). If something else turns out to need access (e.g. \`/dev/fd/N\`), add it explicitly.

Acceptable to ship as-is given the framing, but the precise version is small and worth doing.

### 3. \"Closes #127\" semantics

The PR closes #127, but #127 was scoped to the full unblock — Phase 1 + Phase 2. Closing #127 here means strict mode (which is what users who care about isolation actually want) becomes invisible on the backlog. Two options:

- Change \"Closes #127\" → \"Part of #127\" and leave #127 open until Phase 2 ships
- Close #127 here but open a tracked Phase 2 issue (\`feat(sandbox): implement real strict mode\`) and link it from the README's \"Currently stubbed — tracked in #127\" line

The README currently links #127 as the strict tracking issue. If #127 closes when this PR merges, that link becomes a dead reference to a closed issue. Worth fixing.

### 4. Behavior change in CI and security-sensitive environments

Pre-PR, \`--sandbox auto\` denied external outbound. Post-PR, it allows everything. Anyone relying on the previous behavior in CI (to verify their agent doesn't make unexpected external calls) will see a silent semantic shift.

The PR's \"Notes for reviewers\" mentions this honestly. Worth adding one line to the README/CHANGELOG explicitly: **\"\`auto\` no longer blocks external network. If you relied on the previous behavior, use \`--sandbox strict\` (when it ships) or \`--sandbox off\` + external firewall.\"** Acknowledging the regression is more honest than burying it.

### 5. Minor — design doc cites an unofficial sandbox-exec reference

\`reverse.put.as/wp-content/uploads/2011/09/Apple-Sandbox-Guide-v1.0.pdf\` is a reverse-engineered community doc, not official Apple. Useful (probably the best public reference for \`sandbox-exec\` profile language) but worth labeling as such so readers know the citation isn't authoritative.

## Smaller observations

- Synchronous \`mkdirSync\` in \`BwrapRunner\` constructor is a small code smell, but acceptable here — runner construction happens once per session, the 11 paths are local mkdir calls, and the alternative (lazy-init on first exec) has its own complexity. Fine as-is.
- The Linux tests use \`if (runner.warning) return;\` to skip when bwrap isn't available. Existing pattern, consistent. CI environments without bwrap silently skip the new tests — not a regression, just inheriting the limitation.
- \`~/Library/Caches\` test is macOS-only by directory convention. The profile allow is also macOS-specific. Symmetric.
- Test infrastructure is good — uses \`process.env.HOME ?? homedir()\`, which matches the runtime resolution exactly.

## Verdict

Ship-quality once the two real concerns (sysctl-write justification, /dev narrowing) are addressed. The third (issue housekeeping) is procedural — fix during merge.

The design doc alone justifies the PR — it's the kind of document that prevents the next person from re-litigating these choices. The honest framing of \`auto\` as \"not a security boundary\" is the most important thing in the diff: it sets correct user expectations and clears the path for \`strict\` to be the actual isolation mode without conceptual collision.

Couple of small follow-up issues worth filing if the maintainer wants:
- Phase 2 tracking issue (real \`strict\` implementation)
- Configurable extra-allow list (the design doc deliberately defers this; worth a placeholder issue for when someone needs it)

zjshen14:
## Response to review

Addressed all five points in commit 0d766af. Replying inline:

### 1. ✅ `sysctl-write` removed
Right call — A1 didn't have it, no workflow needs it, and adding capabilities silently is exactly the kind of profile drift we want to avoid. Dropped.

### 2. ✅ `/dev` narrowed to explicit literals
Now allows only `/dev/null`, `/dev/zero`, `/dev/stdout`, `/dev/stderr`, `/dev/tty`, `/dev/urandom`, `/dev/random`. Covers every cited workflow (`curl -o /dev/null`, shell redirects, prompts) without allowing `/dev/disk*`, `/dev/kmsg`, capture devices, etc. Tests still pass.

### 3. ✅ Issue housekeeping — opened Phase 2 tracking issue
Filed [#149 — feat(sandbox): implement real strict mode (Phase 2 of #127)](https://github.com/zjshen14/opencli/issues/149) with the full strict-mode spec from the design doc. Updated the README's strict reference to point to #149 instead of #127, so the link stays live after this PR closes #127.

Kept `Closes #127` in this PR — #127 is fundamentally about the Phase 1 unblock (the user-facing bug that bites every user within minutes). Closing it here is honest about what was fixed.

### 4. ✅ README has an explicit behavior-change note
Added a ⚠ callout under the mode table:
> Prior to this release, `--sandbox auto` denied all external network access. As of #127, `auto` allows external network by default — every real coding workflow (`npm install`, `gh`, `git clone`, `curl`) was blocked otherwise. If you relied on the previous network-deny behavior, use `--sandbox strict` (once #149 lands) or `--sandbox off` plus an external firewall.

Acknowledges the regression instead of burying it.

### 5. ✅ Reference labeled
Design doc now describes the Apple Sandbox Guide PDF as "community-maintained, reverse-engineered — not an official Apple document" with a note that Apple's official `sandbox-exec(1)` man page exists but is sparse.

---

**Smaller observations** — all noted as fine-as-is. Specifically agree that:
- Synchronous `mkdirSync` in `BwrapRunner` constructor is acceptable (one-time, 11 local paths)
- `if (runner.warning) return;` skip pattern on Linux tests is consistent with existing code
- `~/Library/Caches` macOS-only allow is symmetric with the test
- `process.env.HOME ?? homedir()` matches runtime resolution

**Follow-up issues:**
- Phase 2 tracking — done (#149)
- Configurable extra-allow list — can file when someone needs it; design doc deliberately defers

CI should re-run on the new commit. Ready for re-review.

PR #152, opened 2026-05-23
fix(bash): unblock when backgrounded grandchildren hold stdio pipes

### LINKED ISSUE #151: bash tool hangs forever when command backgrounds a long-running process that inherits stdout/stderr pipes
## Repro

In a real session (2026-05-17T22-00-06-184, card_trade project) the agent ran:

\`\`\`bash
kill -9 98812 && rm -rf .next && npm run dev > dev.log 2>&1 & echo \$! > dev.pid
\`\`\`

The bash tool **hung indefinitely** with no \`tool_result\` emitted. After 7+ minutes the user noticed; the OpenCLI session was unrecoverable.

The \`npm run dev\` server itself started successfully and was listening on port 3000 — only the OpenCLI session was stuck.

## Root cause

Bash parses the command as:

\`\`\`bash
(kill -9 98812 && rm -rf .next && npm run dev > dev.log 2>&1) &
echo \$! > dev.pid
\`\`\`

So the entire \`&&\` chain runs inside a backgrounded **subshell**. That subshell:

1. Inherits stdout/stderr **pipe file descriptors** from the spawned \`/bin/sh\` (Node's \`spawn()\` connects to these pipes)
2. Waits for \`npm run dev\` to finish (never — it's a server)
3. Never exits → never closes the pipe FDs → Node's \`spawn.close\` event never fires → bash tool waits forever

The 30-second timeout in [src/tools/exec/sandbox/passthrough.ts](src/tools/exec/sandbox/passthrough.ts) does fire and sends SIGTERM, but only to the immediate \`/bin/sh\` PID. The backgrounded subshell and \`npm run dev\` are in a different process group and the signal doesn't reach them. The pipes stay held open. The Promise never resolves.

**Compare** to the previous working pattern in the same session:

\`\`\`bash
nohup npm run dev > dev.log 2>&1 &
\`\`\`

\`nohup\` explicitly closes stdin/stdout/stderr (replaces with /dev/null and nohup.out), so \`npm run dev\` doesn't inherit any pipes to Node. Backgrounded cleanly, parent shell exits, spawn sees \`close\`.

## Why this is a real bug

The agent has no way to know that an \`&&\`-chained backgrounded command will hang. The LLM correctly emitted a backgrounded \`npm run dev\` — that's a legitimate pattern. The bash tool should either:

- Handle the deadlock (force-resolve when timeout fires)
- Prevent it (process-group spawn + group-kill on timeout)
- Or surface the trap to the LLM via system prompt guidance

Currently the user has to manually kill the OpenCLI process.

## Proposed fix

Two parts:

1. **Process-group spawn + group-kill on timeout** ([src/tools/exec/sandbox/passthrough.ts](src/tools/exec/sandbox/passthrough.ts), \`sandbox-exec.ts\`, \`bwrap.ts\`):
   - Spawn with \`detached: true\` so the child becomes its own process group leader
   - On timeout, send signal to the **process group** via \`process.kill(-proc.pid, signal)\`
   - After SIGTERM, if process still alive after a grace period (~2s), send SIGKILL to the group
   - Force the Promise to resolve with \`exitCode: -1\` and a clear stderr message even if pipes are still open (track timeout separately from close event)

2. **System prompt guidance** ([src/core/prompt.ts](src/core/prompt.ts)):
   - Add explicit guidance: for long-running servers, use \`nohup CMD > log 2>&1 < /dev/null &\`
   - Don't end an \`&&\` chain with a backgrounded long-running process

## Related

- The \`/bin/ps\` and \`/usr/bin/top\` setuid limitations documented in #127 / PR #148
- The agent-stops-mid-todo issue fixed in PR #150

zjshen14:
## Summary

- Live session debugging revealed a hard bash-tool hang when the agent runs a backgrounded long-running process via an \`&&\` chain (e.g. \`kill -9 PID && rm -rf .next && npm run dev > log 2>&1 &\`). The backgrounded subshell holds inherited stdout/stderr pipe FDs forever; \`spawn.close\` never fires; the 30s SIGTERM misses the subshell because it targets only the immediate child PID. Result: the entire OpenCLI session is stuck until the user kills it manually.
- Fix in three parts:
  1. Each sandbox runner (\`passthrough\`, \`sandbox-exec\`, \`bwrap\`) spawns with \`detached: true\`, making the child its own process group leader.
  2. \`spawnAndCollect\` kills the process **group** on timeout (\`process.kill(-proc.pid, sig)\`), escalates SIGTERM → SIGKILL after a 2s grace, then force-resolves the Promise after a 500ms window if \`close\` still hasn't fired. \`proc.stdout/stderr.destroy()\` releases Node's hold on the pipes.
  3. System prompt now tells the agent to use \`nohup CMD > log 2>&1 < /dev/null &\` for long-running servers and to never end an \`&&\` chain with a backgrounded long-running command.

## Related issue

Closes #151

## Test plan

- [x] \`npm run typecheck && npm run lint && npm run format:check && npm test\` — 550 tests pass
- [x] New regression test: \`PassthroughRunner > unblocks within grace period when command backgrounds a process holding stdio\` — runs \`sleep 30 & echo started\` with a 300ms timeout, asserts the call resolves in under 5s. Old behavior would hang ~30s.
- [x] Existing timeout test still passes (\`returns exitCode -1 on timeout\` — \`sleep 60\` with timeout=100ms)
- [ ] Manual verification: from a fresh REPL, ask the agent to run \`npm run dev > dev.log 2>&1 & sleep 2 && curl localhost:3000\` and confirm it returns promptly with the curl output

## Notes for reviewers

The three-stage timeout escalation (SIGTERM → grace → SIGKILL → grace → force-resolve) is overkill for most cases but handles the worst case: a backgrounded grandchild that ignores both signals because it's in a different process group AND has inherited pipe FDs. The two grace windows (2000ms + 500ms) are conservative; could be tightened later if telemetry shows shorter is sufficient.

\`detached: true\` is safe here because we still call \`spawn.on(\"close\", ...)\` — Node tracks the child until close fires. We don't \`unref()\` (which would let Node exit even if the child is still running). The only behavior change is signal targeting: signals sent via \`process.kill(-proc.pid, ...)\` now hit the whole group.

The system prompt addition is paired with the runtime fix because even with the runtime fix, the agent's command would now hit the 30s timeout (still slow). The prompt nudge teaches the agent to use \`nohup\` upfront, avoiding the 30s wait entirely.

### DIFF
--- a/src/core/prompt.ts
+++ b/src/core/prompt.ts
@@ -221,7 +221,7 @@ After every code change:
 - One \`edit\` call per file per turn to avoid conflicts
 - **edit**: always \`read\` the file first; \`old_string\` must match exactly — whitespace and indentation included; if it fails with "not found", re-read and try again with the exact content
 - **bash**: if a command fails, read the full error output before retrying; never retry unchanged
-- **bash long-running servers**: use \`nohup CMD > log 2>&1 < /dev/null &\` to background dev servers and daemons. Never end an \`&&\` chain with a backgrounded long-running command (e.g. \`A && B && server &\`) — the backgrounded subshell inherits stdio pipes and the call will hang until timeout. After starting, verify with \`sleep 2 && curl -s localhost:PORT\` or \`tail log\`.
+- **bash long-running servers**: use \`nohup CMD > log 2>&1 < /dev/null &\` to background dev servers and daemons — all three FDs must be redirected so the shell can return immediately. Never end an \`&&\` chain with a backgrounded long-running command (e.g. \`A && B && server &\`) — the backgrounded subshell inherits stdio pipes and the call will hang until timeout. After starting, verify with \`sleep 2 && curl -s localhost:PORT\` or \`tail log\`.
 - **think**: use before starting any change that touches more than two files; reason through the approach and order of changes
 - **todo_write**: for tasks with more than three steps, write the steps first and check them off as you go
 
--- a/src/core/prompt.ts
+++ b/src/core/prompt.ts
@@ -221,6 +221,7 @@ After every code change:
 - One \`edit\` call per file per turn to avoid conflicts
 - **edit**: always \`read\` the file first; \`old_string\` must match exactly — whitespace and indentation included; if it fails with "not found", re-read and try again with the exact content
 - **bash**: if a command fails, read the full error output before retrying; never retry unchanged
+- **bash long-running servers**: use \`nohup CMD > log 2>&1 < /dev/null &\` to background dev servers and daemons. Never end an \`&&\` chain with a backgrounded long-running command (e.g. \`A && B && server &\`) — the backgrounded subshell inherits stdio pipes and the call will hang until timeout. After starting, verify with \`sleep 2 && curl -s localhost:PORT\` or \`tail log\`.
 - **think**: use before starting any change that touches more than two files; reason through the approach and order of changes
 - **todo_write**: for tasks with more than three steps, write the steps first and check them off as you go
 
--- a/src/tools/exec/sandbox/bwrap.ts
+++ b/src/tools/exec/sandbox/bwrap.ts
@@ -134,6 +134,10 @@ export class BwrapRunner implements SandboxRunner {
         cwd: opts.cwd,
         env: opts.env ?? process.env,
         stdio: ["ignore", "pipe", "pipe"],
+        // bwrap becomes its own process group leader so spawnAndCollect
+        // can kill the whole group (including backgrounded grandchildren)
+        // on timeout via process.kill(-proc.pid, ...).
+        detached: true,
       },
     );
 
--- a/src/tools/exec/sandbox/passthrough.test.ts
+++ b/src/tools/exec/sandbox/passthrough.test.ts
@@ -25,12 +25,19 @@ describe("PassthroughRunner", () => {
     expect(result.exitCode).toBe(-1);
   });
 
-  it("unblocks within grace period when command backgrounds a process holding stdio", async () => {
+  it("unblocks promptly when command backgrounds a process holding stdio", async () => {
     // Issue #151 repro: a backgrounded grandchild that doesn't redirect
     // stdin/stdout/stderr keeps the spawn's pipe FDs open. Before the fix,
-    // the bash tool's Promise never resolved because `close` waited on those
-    // pipes. With detached spawn + process-group kill on timeout, the tool
-    // must unblock within (timeout + grace + force-resolve buffer).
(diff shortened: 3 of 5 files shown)


zjshen14:
## Response to the second review — alignment check

The two reviews agree on 2 of 6 points, and the second reviewer corrected a real error in my self-review. Addressed in commit 287a5a9.

### Where we agree (both reviewers flagged it)

**1. Timer init hack** — both reviewers called out the dummy-setTimeout pattern. Fixed:
\`\`\`typescript
let killTimer: NodeJS.Timeout | undefined;
let forceTimer: NodeJS.Timeout | undefined;
\`\`\`
\`clearTimeout(undefined)\` is a documented no-op, so finish() needs no conditional check.

**4. Test assertion too permissive** — both reviewers flagged \`expect([0, -1]).toContain(...)\`. The second reviewer was *correct* that exitCode is deterministically -1; my self-review claim that \`detached: true\` would trigger a SIGHUP cascade was wrong. Without a controlling terminal (stdio is piped), session-leader-exit does NOT auto-send SIGHUP. The timer is what kills sleep. Verified empirically — my "tighten to 0" attempt failed at 5.12s with exitCode -1. Corrected to:
\`\`\`typescript
timeout: 300,
// expected: ~300-500ms real
expect(elapsed).toBeLessThan(2000);
expect(result.exitCode).toBe(-1);
expect(result.stdout).toContain("started");
\`\`\`

### Where the second reviewer made points I hadn't covered

**Prompt clarity (#6)** — added rationale that all three FDs must be redirected so the shell can return. Without this, an agent might substitute \`nohup\` into an \`&&\` chain without understanding why redirection matters.

**Defensive pid check** — already in my self-review (\`<= 0\` guard). Fixed independently.

### Where I disagree or am deferring

**SIGINT note (#2)** — agreed, the \`stdio: [\"ignore\", ...]\` means Ctrl+C isn't expected to reach the child anyway. Not adding to PR description; could land in architecture docs later.

**bwrap PID namespace concern (#3)** — our \`bwrap.ts\` does NOT use \`--unshare-pid\`, so the host can see and signal the sandboxed process. \`process.kill(-bwrap_pid, ...)\` targets bwrap's process group which contains the sandboxed \`/bin/sh\` and any backgrounded children. The fix works. Not adding a Linux smoke test because the platform-conditional ceremony outweighs the value — the shared spawnAndCollect logic is the actual code under test.

**Telemetry on force-resolve (#5)** — fair point but defer. The path is rare enough that I'd rather wait to see if it shows up in real sessions before adding observability surface. If #151 reoccurs, this is the first thing to add.

### Status

550 tests pass. Branch updated. Ready for re-review and merge.

PR #153, opened 2026-05-24
fix(sandbox): kill process group on timeout to prevent pipe-hang deadlock

### LINKED ISSUE #151: bash tool hangs forever when command backgrounds a long-running process that inherits stdout/stderr pipes

zjshen14:
## Summary

- **Root cause**: when an agent command backgrounds a long-running process via an `&&`-chain (e.g. `kill -9 PID && npm run dev > dev.log 2>&1 &`), bash runs the whole chain in a backgrounded subshell that **inherits Node's stdout/stderr pipe file descriptors**. The server never exits, never closes those FDs, so Node's `spawn('close')` event never fires and the Promise hangs forever — making the session unrecoverable.
- **Process-group kill**: all three runners (`PassthroughRunner`, `SandboxExecRunner`, `BwrapRunner`) now spawn with `detached: true` so the child shell becomes its own process group leader. `spawnAndCollect` now sends `SIGTERM` to the **entire group** (`process.kill(-proc.pid, "SIGTERM")`) on timeout, reaching backgrounded children that previously survived.
- **Force-resolve after grace period**: after SIGTERM, a 2-second grace timer fires `SIGKILL` on the group and then force-resolves the Promise (destroying the stdout/stderr streams to release pipe FD references) — so the agent loop recovers even if some process escapes the kill.
- **System-prompt guidance**: added a `bash (background servers)` bullet to `DEFAULT_SYSTEM_INSTRUCTION` directing the LLM to use `nohup CMD > log 2>&1 < /dev/null &` for long-running servers instead of ending an `&&`-chain with a backgrounded process.

Closes #151

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — all 539 tests pass
- [x] New regression test `PassthroughRunner > resolves after timeout when a backgrounded child holds pipes open`: runs `bash -c "sleep 60 &"` with a 300 ms timeout; verifies `exitCode === -1` and completes in < 5 s (actual: ~303 ms — SIGTERM reaches the group and sleep exits cleanly)
- [x] Existing timeout test (`sleep 60` with 100 ms timeout) still passes unchanged

https://claude.ai/code/session_019Lo1p8X6xUnqCvR4AKa8DX

---
_Generated by [Claude Code](https://claude.ai/code/session_019Lo1p8X6xUnqCvR4AKa8DX)_

### DIFF
--- a/src/core/prompt.ts
+++ b/src/core/prompt.ts
@@ -221,6 +221,7 @@ After every code change:
 - One \`edit\` call per file per turn to avoid conflicts
 - **edit**: always \`read\` the file first; \`old_string\` must match exactly — whitespace and indentation included; if it fails with "not found", re-read and try again with the exact content
 - **bash**: if a command fails, read the full error output before retrying; never retry unchanged
+- **bash (background servers)**: use \`nohup CMD > log 2>&1 < /dev/null &\` to start long-running processes; never end an \`&&\` chain with a backgrounded server — the subshell inherits pipe file descriptors and the tool call will hang until the server exits
 - **think**: use before starting any change that touches more than two files; reason through the approach and order of changes
 - **todo_write**: for tasks with more than three steps, write the steps first and check them off as you go
 
--- a/src/tools/exec/sandbox/bwrap.ts
+++ b/src/tools/exec/sandbox/bwrap.ts
@@ -134,6 +134,7 @@ export class BwrapRunner implements SandboxRunner {
         cwd: opts.cwd,
         env: opts.env ?? process.env,
         stdio: ["ignore", "pipe", "pipe"],
+        detached: true,
       },
     );
 
--- a/src/tools/exec/sandbox/passthrough.test.ts
+++ b/src/tools/exec/sandbox/passthrough.test.ts
@@ -25,6 +25,16 @@ describe("PassthroughRunner", () => {
     expect(result.exitCode).toBe(-1);
   });
 
+  it("resolves after timeout when a backgrounded child holds pipes open", async () => {
+    // Reproduces: bash runs `sleep 60 &` — shell exits immediately but sleep
+    // inherits the pipe FDs, so close() never fires without a group kill.
+    const start = Date.now();
+    const result = await runner.exec("sleep 60 &", { cwd: process.cwd(), timeout: 300 });
+    expect(result.exitCode).toBe(-1);
+    // Must resolve well within timeout + 2 s grace, not hang indefinitely.
+    expect(Date.now() - start).toBeLessThan(5_000);
+  });
+
   it("exposes mode and null warning", () => {
     expect(runner.mode).toBe("off");
     expect(runner.warning).toBeNull();
--- a/src/tools/exec/sandbox/passthrough.ts
+++ b/src/tools/exec/sandbox/passthrough.ts
@@ -1,10 +1,6 @@
 import { spawn } from "node:child_process";
 import type { SandboxExecOptions, SandboxExecResult, SandboxMode, SandboxRunner } from "./types.js";
 
-/**
- * Collect stdout/stderr from a spawned process and resolve when it closes.
- * Sends SIGTERM after timeoutMs and resolves with exitCode -1.
- */
 export async function spawnAndCollect(
   proc: ReturnType<typeof spawn>,
   timeoutMs: number,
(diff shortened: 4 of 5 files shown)


zjshen14:
The core fix is correct: `detached: true` + negative-PID group kill is the right pattern for cleaning up pipe-holding background children, and the two-phase SIGTERM → SIGKILL with an unref'd timer handles the race between timeout and normal exit cleanly. The regression test (`sleep 60 &` resolving within 5s) directly targets the root cause.

Two things to flag before landing:

**Ordering conflict with PR #155**
Both this PR and #155 modify `src/tools/exec/sandbox/bwrap.ts` (this PR adds `detached: true`; #155 refactors the entire spawn call block). The `mergeable_state: unknown` on this PR may already reflect that conflict. This PR should land first since #155 appears to have been built with `detached: true` already present. If the order is reversed, this PR will conflict.

**Defensive guard for `proc.pid` in `killGroup`**
`killGroup` does `process.kill(-proc.pid!, sig)` with a non-null assertion. If `spawn` fails to assign a PID (which Node.js permits in error cases), this throws rather than fails gracefully. A `if (!proc.pid) return;` guard at the top of `killGroup` would be cheap and defensive.

Cannot merge directly (touches `src/tools/exec/sandbox/`). Recommend landing before #155 to avoid conflicts.



---

zjshen14:
Closing — superseded by PR #152 which landed on main at commit 90e963a and already closed #151. The DIRTY merge state confirms the changes overlap with the already-merged process-group-kill + force-resolve fix.

Issue #34, opened 2026-05-01
Feature: Human-in-the-loop tool use confirmation


zjshen14:
## Overview

OpenCLI currently handles risky tool use through a coarse-grained regex blocklist in the bash tool and a read-only plan mode. There is no interactive confirmation flow — dangerous commands are either silently blocked or allowed with no user input. This issue proposes a proper human-in-the-loop (HITL) confirmation system.

---

## Current State

**What exists today:**

- **Hard-blocked regex patterns** in `src/tools/exec/bash.ts` — 7 patterns (`rm -rf`, `git push --force`, etc.) that reject with an error message. No user override possible.
- **Plan mode** (`/plan`) in `src/cli/repl.ts` — run the whole agent in read-only mode, then approve/edit/cancel the plan before execution. Coarse-grained: all-or-nothing per plan.
- **Executor-level `readOnly` flag** in `src/agent/executor.ts` — blocks write tools at runtime as a second layer in plan mode.
- **System prompt guidance** in `src/agent/prompt.ts` — tells the model to warn before destructive operations. Prompt-level only, not enforced.
- **Event-driven reminders** — appended to tool results (e.g. "never commit without explicit user request"), but non-blocking.

**What's missing:**

- No per-tool-call confirmation dialog (y/n/always/never)
- No persistent allow-list that survives across sessions
- No configurable permission rules per project
- No distinction between soft-block (ask) and hard-block (deny)

---

### Claude Code (Anthropic)

Claude Code has the most complete implementation in this space. Key design:

**Three-tier rule precedence** (evaluated in order, first match wins):
1. `deny` — unconditionally block
2. `ask` — prompt for confirmation
3. `allow` — auto-approve

Rules are defined in `.claude/settings.json` at project level or `~/.claude/settings.json` at user level (shareable via git):

```json
{
  "permissions": {
    "deny": ["Bash(rm -rf /)"],
    "ask": ["Bash(git push *)"],
    "allow": ["Bash(npm run *)", "Bash(git commit *)"]
  }
}
```

**Granularity**: Wildcard matching on bash commands (with word-boundary-aware parsing). Compound commands (`cmd1 && cmd2`) are split and each subcommand checked independently.

**Session vs. persistent**: File-edit approvals last until session end; bash command approvals are saved permanently to `settings.json` when the user picks "don't ask again."

[...]

PR #156, opened 2026-05-25
feat(permissions): add configurable deny rules to the confirmation system

zjshen14:
## Summary

- `Permissions` had only an `allow` list; `deny` was anticipated by `settings.test.ts` (which already verified deny entries survive a round-trip through `saveSettings`) but never enforced
- Adds `deny?: string[]` to the `Permissions` type in `config.ts` — works in both `~/.opencli/config.json` (global) and `.opencli/settings.json` (project-level)
- Adds `globMatch` and `matchesDenyPattern` helpers (exported for testing): pattern format is `toolName(argGlob)` where `*` matches any chars; primary arg is `args.command` for bash, `args.file_path` for write/edit, `JSON.stringify(args)` otherwise
- `createConfirmFn` merges global + project deny lists and checks them **before** the allow list — a matching tool call is rejected without prompting even if it also appears in allow

Example config:
```json
{
  "permissions": {
    "deny": ["bash(rm -rf *)", "bash(git push --force*)"],
    "allow": ["bash(npm test)"]
  }
}
```

Note: deny rules apply to tools that already call `requiresConfirmation` (non-SAFE_COMMANDS for bash, out-of-cwd writes). Adding `ask` rules that force confirmation for SAFE_COMMANDS requires executor-layer changes and is deferred.

## Related issue

Part of #34

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — all pass (562 tests, 14 new in `src/cli/confirm.test.ts`)
- [x] New tests cover: exact patterns, `*` glob, cross-tool no-match, bash/write/edit primary-arg selection, malformed patterns skipped, deny takes precedence over allow, multi-pattern list

https://claude.ai/code/session_019spv4KZ2oRudPf1SA4PPaY

---
_Generated by [Claude Code](https://claude.ai/code/session_019spv4KZ2oRudPf1SA4PPaY)_

### DIFF
--- a/src/cli/confirm.test.ts
+++ b/src/cli/confirm.test.ts
@@ -1,0 +1,85 @@
+import { describe, it, expect } from "vitest";
+import { globMatch, matchesDenyPattern } from "./confirm.js";
+
+describe("globMatch", () => {
+  it("matches an exact string with no wildcards", () => {
+    expect(globMatch("git status", "git status")).toBe(true);
+    expect(globMatch("git status", "git status --short")).toBe(false);
+  });
+
+  it("matches * as zero or more characters", () => {
+    expect(globMatch("rm -rf *", "rm -rf /tmp")).toBe(true);
+    expect(globMatch("rm -rf *", "rm -rf .")).toBe(true);
+    expect(globMatch("rm -rf *", "rm -rf")).toBe(false); // trailing space required
+    expect(globMatch("rm -rf*", "rm -rf")).toBe(true); // no space → * matches empty
+  });
+
+  it("escapes regex special chars in the pattern literally", () => {
+    expect(globMatch("npm run test", "npm run test")).toBe(true);
+    expect(globMatch("file.ts", "file.ts")).toBe(true);
+    expect(globMatch("file.ts", "fileXts")).toBe(false); // dot is literal
+  });
+
+  it("supports leading wildcard", () => {
+    expect(globMatch("*.ts", "foo.ts")).toBe(true);
+    expect(globMatch("*.ts", "foo.js")).toBe(false);
+  });
+});
+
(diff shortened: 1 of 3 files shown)


zjshen14:
The `matchesDenyPattern` + `globMatch` design is clean — deny-before-allow priority is correct, the pattern syntax is intuitive, and the 14 tests cover all the interesting cases well. A few things worth looking at before landing:

**`src/cli/confirm.ts` — glob syntax is underdocumented**
`*` translates to `.*` which is correct for single-level wildcards. However `**` would expand to `.*.*` (which happens to behave identically to `.*` in most cases, but is not path-separator-aware), and `?` is unsupported and passed through literally. Since this is a new user-visible API surface, it's worth adding a short doc note on exactly which syntax is supported so users who expect POSIX glob semantics aren't surprised.

**`src/cli/confirm.ts` — `JSON.stringify` fallback is opaque for users**
For any tool other than `bash`, `write`, and `edit`, the pattern arg is matched against `JSON.stringify(args)`. A user who writes `read(src/config.ts)` will never match — they'd need `read({"file_path":"src/config.ts"})`. This is probably intentional for now, but the feature docs and/or an inline comment should call it out. A test case demonstrating this behavior explicitly would also help future contributors understand the contract.

**Minor: closing-paren in bash args**
`pattern.slice(parenOpen + 1, -1)` extracts the arg between the first `(` and the last character. If a bash command legitimately contains a closing paren — e.g. `bash(echo "done (ok)")` — the extracted arg becomes `echo "done (ok)"` minus its final `)`, yielding `echo "done (ok"`. Very unlikely in practice but worth a note or a smarter slice (find the *last* `)` as the close delimiter).

Recommendation: merge after the glob-syntax documentation gap is addressed. The logic itself is correct.



---

PR #164, opened 2026-05-28
feat(permissions): add ask rules to force confirmation for SAFE_COMMANDS

### LINKED ISSUE #34: Feature: Human-in-the-loop tool use confirmation

zjshen14:
## Summary

- Adds `ask?: string[]` to the `Permissions` type in both `config.ts` and `settings.ts`, completing the deny side of Phase 2 (#34) with an ask-side counterpart
- Users can now list `"toolName(argGlob)"` patterns under `permissions.ask` in `.opencli/settings.json` (project-scoped) or `~/.opencli/config.json` (global) to force the HITL confirmation dialog even for commands the tool itself considers safe — e.g., forcing a prompt before any `git push`
- Exports `createForcesConfirmationFn(askPatterns)` from `confirm.ts` — a pure synchronous function, easily unit-tested without mocks
- Adds `ExecutorDeps.forcesConfirmation?` and gates confirmation on `requiresConfirmation || forcesConfirmation`
- Adds `Agent.setForcesConfirmationFn()` to wire the function through to the executor; `createConfirmFn` returns `{ confirmFn, forcesConfirmation }` and both are set from `repl.ts` and `cli/index.ts`
- Evaluation order: deny → (allow short-circuits ask) → ask → default tool behaviour

## Related issue

Closes #34 (Part of Phase 2 — ask rules; Phase 3 AI-based classification is a separate stretch goal)

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — all pass (601 tests, 0 failures)
- [x] 5 new unit tests for `createForcesConfirmationFn` covering: empty patterns, bash match, write match, no tool match, multi-pattern merge
- [x] 3 new executor tests: forces confirmation when `requiresConfirmation` is false, denies when `forcesConfirmation` triggers and `confirmFn` denies, skips `confirmFn` when both return false

---
_Generated by [Claude Code](https://claude.ai/code/session_0141SHjwe6zcA1AdeugsYpwR)_

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,7 +70,7 @@ src/
   state/
     config.ts       # ~/.opencli/config.json load/save; exports AGENT_DIR, Config (incl. anthropicApiKey)
     session.ts      # JSONL session logs at ~/.opencli/projects/<cwd>/; create, list, resume
-    settings.ts     # .opencli/settings.json load/save; holds project-level Permissions (HITL allow/deny list)
+    settings.ts     # .opencli/settings.json load/save; holds project-level Permissions (HITL allow/ask/deny list)
     snapshot.ts     # SnapshotManager — git stash create/restore for per-session working-tree snapshots
 ```
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -71,7 +71,7 @@ src/
   state/
     config.ts       # ~/.opencli/config.json load/save; exports AGENT_DIR, Config (incl. anthropicApiKey)
     session.ts      # JSONL session logs at ~/.opencli/projects/<cwd>/; create, list, resume
-    settings.ts     # .opencli/settings.json load/save; holds project-level Permissions (HITL allow/deny list)
+    settings.ts     # .opencli/settings.json load/save; holds project-level Permissions (HITL allow/ask/deny list)
     snapshot.ts     # SnapshotManager — git stash create/restore for per-session working-tree snapshots
 ```
 
--- a/src/cli/confirm.test.ts
+++ b/src/cli/confirm.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { globMatch, matchesDenyPattern } from "./confirm.js";
+import { globMatch, matchesDenyPattern, createForcesConfirmationFn } from "./confirm.js";
 
 describe("globMatch", () => {
   it("matches an exact string with no wildcards", () => {
@@ -81,5 +81,38 @@ describe("matchesDenyPattern", () => {
     const patterns = ['read({"file_path":"secret.txt"})'];
     expect(matchesDenyPattern(patterns, "read", { file_path: "secret.txt" })).toBe(true);
     expect(matchesDenyPattern(patterns, "read", { file_path: "other.txt" })).toBe(false);
+  });
+});
+
+describe("createForcesConfirmationFn", () => {
+  it("returns false for all calls when ask patterns list is empty", () => {
+    const fn = createForcesConfirmationFn([]);
+    expect(fn("bash", { command: "git status" })).toBe(false);
+    expect(fn("write", { file_path: "src/foo.ts" })).toBe(false);
+  });
+
+  it("returns true when bash command matches an ask pattern", () => {
+    const fn = createForcesConfirmationFn(["bash(git push*)"]);
+    expect(fn("bash", { command: "git push origin main" })).toBe(true);
+    expect(fn("bash", { command: "git status" })).toBe(false);
+  });
+
+  it("returns true when write path matches an ask pattern", () => {
+    const fn = createForcesConfirmationFn(["write(src/*)"]);
+    expect(fn("write", { file_path: "src/index.ts" })).toBe(true);
+    expect(fn("write", { file_path: "test/index.ts" })).toBe(false);
+  });
+
+  it("returns false for a non-matching tool name", () => {
+    const fn = createForcesConfirmationFn(["bash(git push*)"]);
+    expect(fn("write", { command: "git push origin main" })).toBe(false);
+  });
+
+  it("matches across multiple patterns", () => {
+    const fn = createForcesConfirmationFn(["bash(git push*)", "write(src/*)"]);
+    expect(fn("bash", { command: "git push origin main" })).toBe(true);
+    expect(fn("write", { file_path: "src/index.ts" })).toBe(true);
+    expect(fn("bash", { command: "git status" })).toBe(false);
+    expect(fn("write", { file_path: "test/index.ts" })).toBe(false);
   });
 });
--- a/src/cli/confirm.ts
+++ b/src/cli/confirm.ts
@@ -39,7 +39,26 @@ export function matchesDenyPattern(
   return false;
 }
 
-export async function createConfirmFn(): Promise<ConfirmFn> {
+/** Builds a sync function that returns true when a tool call matches one of the
+ *  given ask patterns and must therefore be confirmed even if the tool's own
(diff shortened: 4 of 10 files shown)


zjshen14:
Solid implementation with thorough tests. A small note on the PR description: it says `ask?` was added to "both `config.ts` and `settings.ts`" — `settings.ts` doesn't define its own `Permissions`, it imports the type from `config.ts`, so the single change to `config.ts` is sufficient and the behavior is exactly right. The evaluation order works as described: `forcesConfirmation` gates entry into `confirmFn`, where `deny` patterns and `allow` short-circuits remain authoritative in the correct priority order. The three new executor tests (forces confirm, denies on deny, skips when both false) cover the branching logic precisely. CI is green.

One edge to be aware of: when `process.stdin.isTTY` is false, `confirmFn` returns `"deny"` immediately, so `ask` patterns effectively behave as `deny` in non-interactive/headless mode. This is consistent with the existing security-first design but worth noting in the `ask` documentation if one is written.

Merging.



---

Issue #43, opened 2026-05-03
Research: features to learn from OpenHands, Open Interpreter, and OpenCode


zjshen14:
## Overview

Competitive research across three open-source coding agents to identify features worth adopting in OpenCLI. Findings are grouped by source tool, then by implementation priority.

---

## OpenHands (All-Hands-AI/OpenHands)

OpenHands' thesis is **event sourcing + optional isolation = production-ready agents**. It solves long-horizon problems that CLI agents mostly sidestep.

### Architecture

- **Event-sourced state**: every interaction is an immutable event appended to an `EventLog`. Enables deterministic replay, cost analysis, and fault recovery without special-case logic.
- **Three-component model**: `Agent` (stateless, emits Actions from history) + `Conversation` (stateful, drives loop, persists EventLog) + `Workspace` (executes Actions, returns Observations).
- **Workspace abstraction**: `LocalWorkspace` / `DockerWorkspace` / `RemoteWorkspace` are swappable — the same agent code runs locally or in a container without modification.

### Stuck Detection (5 semantic patterns)

OpenHands detects loops semantically (ignoring timestamps/IDs), not just by retry count:

1. **Repeating cycles** — same action-observation pair 4+ times
2. **Error loops** — identical action fails 3+ times
3. **Agent monologues** — 3+ consecutive model messages without user input
4. **Alternating ping-pong** — two action-observation pairs alternate 6+ times
5. **Context window errors** — memory pressure triggers its own recovery path

OpenCLI currently aborts after 3 identical consecutive call signatures (`core.ts:STUCK_THRESHOLD`). Adopting patterns 2–5 would catch loops the current check misses.

### Context Compression

Pluggable condenser strategies (relevant to issue #26):

| Strategy | Behaviour |
|---|---|
| `NoOpCondenser` | No compression (testing) |
| `LLMSummarizingCondenser` | Summarises the middle of the event log when event count > threshold (default: 120) |
| `LLMAttentionCondenser` | Ranks events by importance, keeps top-N without generating a summary |
| `PipelineCondenser` | Chains multiple condensers sequentially |

**Key differentiator**: the model can call a `request_condensation` tool itself when it senses context pressure — rather than waiting for a hard external threshold. Result: ~2× per-turn cost reduction after compaction activates; 54% SWE-bench solve rate maintained.

### Skill System

Two skill types not present in OpenCLI:

[...]

PR #165, opened 2026-05-28
feat: add /undo command to remove the last agent turn from history

zjshen14:
## Summary

- Adds `/undo` slash command to the REPL that removes the last user message plus all subsequent agent messages (including any tool call round-trips) from conversation history
- Lets users recover from a wrong prompt or a bad agent response without clearing the entire session
- Implements `ContextManager.popTurn()` (scans backwards for the last clean user text message, truncates from that index) and exposes it via `Agent.undoLastTurn()`

## Related issue

Part of #43 (quick win: `/undo` command from the "Priority Feature Backlog")

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — all pass (599 tests, 0 failures)
- [x] 6 new unit tests for `ContextManager.popTurn()` covering: empty history, single user message, user+model pair, full tool-call round-trip, preserving earlier turns, and repeated undo calls

---
_Generated by [Claude Code](https://claude.ai/code/session_01AwCpYYQZJm98S3u1utYxfw)_

### DIFF
--- a/src/cli/repl.ts
+++ b/src/cli/repl.ts
@@ -22,6 +22,7 @@ const BUILTIN_COMMANDS: SlashCommand[] = [
   { name: "compact", description: "summarize older conversation history to free context" },
   { name: "context", description: "show current token usage vs. context window" },
   { name: "rewind", description: "undo agent file changes since last snapshot" },
+  { name: "undo", description: "remove the last user message and agent response from history" },
   { name: "clear", description: "clear conversation history" },
   { name: "exit", description: "exit the agent" },
 ];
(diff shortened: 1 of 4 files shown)


zjshen14:
Clean, minimal implementation. `popTurn()` correctly identifies the last clean user message by scanning backwards for the first `role: "user"` entry without `function_result` parts — tool round-trips are handled naturally because those intermediate user messages carry `function_result` parts. The 6-test suite covers the meaningful edge cases including repeated undo. CI is green.

One thing worth tracking on the issue: the in-memory history is trimmed but session JSONL logs are append-only, so `/undo` has no effect on the log — a session resumed after an undo will replay the undone turn from the log. Acceptable limitation for this first-pass implementation, but worth documenting on #43 so it's not surprising to future contributors.

Merging.



---

Issue #89, opened 2026-05-10
feat: Phase A6 — inline diffs + compact tool rows in renderer


zjshen14:
## Overview

Phase A6 in the [roadmap](docs/roadmap.md): renderer-only changes that improve the visual density of tool output in the REPL.

**Inline diffs** — when the `edit` tool runs, render a compact colored diff (like `git diff --stat` + hunk preview) instead of showing the full old/new strings.

**Compact tool rows** — collapse tool call + result into a single indented line when the result is short (e.g. `✓ read src/foo.ts (42 lines)`). Only expand to the current multi-line format for large outputs or errors.

These are purely renderer changes in `src/cli/renderer.ts`. No logic changes to agent, executor, or tools.

## Acceptance criteria

- [ ] `edit` tool shows a colored unified diff (context: 3 lines, additions green, deletions red)
- [ ] Short tool results render on a single compact line
- [ ] Long outputs (>5 lines) still render expanded
- [ ] Error results always render expanded
- [ ] `npm run typecheck && npm run lint && npm test` pass

zjshen14:
Implemented and merged in PR #92 (squash commit `0578388`).

**What changed** (`src/cli/renderer.ts`):
- Added `printToolResultExpanded()`: shows actual output lines (up to `MAX_EXPANDED_LINES = 20`) with a header. Errors render with `✗` in red + full error body; long outputs render with `✓` + line-count header + all lines up to the cap (with an overflow note).
- Updated `printToolResult()` to dispatch to `printToolResultExpanded` when the result has >5 lines or starts with `Error:`. Short (≤5 lines, non-error) results keep the existing compact `✓` summary line unchanged.
- All COMPACT_TOOLS one-liner rendering and inline edit diffs (`printEditDiff`) were already present.

**Also fixed**: `snapshot.test.ts` — `initRepo()` now sets `commit.gpgsign=false` in the temp test repo so snapshot tests pass in environments where the global git config requires signing.

**Tests**: 9 new tests added; all 443 tests pass.

---
_Generated by [Claude Code](https://claude.ai/code)_

zjshen14:
Reviewed this issue as part of an autonomous agent run. The inline diff and compact tool row features appear fully implemented in `src/cli/renderer.ts` (`printEditDiff`, `COMPACT_TOOLS`, `printToolResultExpanded`) from the merged PR #92.

[...]

PR #183, opened 2026-05-30
feat(input): multi-line input with backslash + Enter continuation

zjshen14:
## Summary

Closes Tier-1 item #6 in [docs/cli-ux-research.md](docs/cli-ux-research.md) — multi-line prompt composition via \`\\\` + Enter continuation. Previously, building a multi-line prompt relied on whatever the terminal happened to do; this teaches OpenCLI's raw-mode reader to handle it explicitly.

## Behavior

- Typing a literal \`\\\` at the end of a line then pressing Enter strips the backslash and inserts a newline. The new line is rendered with a dim \`· \` continuation prompt that visually distinguishes it from the green \`› \` first-line prompt.
- Bare Enter (no trailing \`\\\`) submits as before. The full buffer including embedded newlines is returned from \`readLine()\`.
- Backspace at column 0 of a continuation line joins it back into the previous line — \`deleteBeforeCursor\` already handles this since it just deletes the inter-line \`\\n\` character.
- The slash-command popup is suppressed once the input contains a \`\\n\` — slash completion is only meaningful on the first line.
- Up/Down arrows still navigate history (no in-buffer line navigation in this MVP — see *Notes for reviewers*).

## Visual

\`\`\`
› first line
· second line
· third line
\`\`\`

## Related issue

Closes the multi-line-input gap in #89 (A6 — inline diffs + compact tool rows). The companion #182 (just merged) covers the other remaining A6 acceptance-criterion gap.

## Test plan

- [x] \`npm run typecheck && npm run lint && npm run format:check && npm test\` — 631 tests pass (+9 new for \`cursorLineCol\`)
- [x] Unit tests for \`cursorLineCol\` cover: single-line cursor positions, the off-by-one edge between end-of-line-N and start-of-line-N+1, three-line input, clamp on overflow, empty input
- [ ] Manual smoke-test the full interactive flow:
  - \`npm run build && npm install -g .\`
  - \`opencli chat\`
  - Type \`Read this file:\\\` then Enter — cursor drops to a \`· \` continuation prompt
  - Type \`./src/foo.ts\` and press Enter — agent receives \`Read this file:\\n./src/foo.ts\`
  - Backspace at the start of the continuation line collapses the lines

## Notes for reviewers

The render path stays compatible with the existing single-line case: when input has no \`\\n\`, the math reduces to the previous behavior. \`cursorRowsBelowTop\` is the one new piece of state — it just tracks how many lines down from the prompt row the cursor ended at the previous render, so the next render can back up to the top of the area before clearing.

In-buffer Up/Down navigation across lines is intentionally not added — that's a larger change touching history-vs-buffer disambiguation. Users can edit a multi-line buffer by Backspacing to the line they want; if that gets annoying in practice we do the bigger arrow rework as a follow-up.

The render() integration isn't directly unit-tested because the raw-mode keypress path has no test fixture today. The tricky math (cursor offset → (line, col)) is fully covered by tests on the extracted \`cursorLineCol\` helper.

### DIFF
--- a/src/cli/input.test.ts
+++ b/src/cli/input.test.ts
@@ -20,6 +20,7 @@ const {
   deleteBeforeCursor,
   deleteWordBeforeCursor,
   renderSelectOptions,
+  cursorLineCol,
 } = await import("./input.js");
 
 describe("insertAtCursor", () => {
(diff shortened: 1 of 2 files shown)


zjshen14:
## Review

### Overview
Teaches `readLine` to handle `\`+Enter as line continuation: strips the trailing backslash, inserts `\n`, and re-renders prompt + multi-line input with a dim `· ` continuation marker. The cursor-positioning math is extracted into a new `cursorLineCol` helper with thorough unit coverage.

### Strengths
- `cursorLineCol` is small, pure, and has 7 well-chosen unit tests (boundary at the `\n`, three-line, empty, overflow clamp). The tricky cursor math is exactly where the testable seam lives — good factoring.
- Render path stays compatible with single-line input: when `lines.length === 1` and there's no popup, `cursorRowsBelowTop` stays 0 and the math reduces to the old behavior. Diff-friendly.
- Popup is correctly suppressed for multi-line input via `lines.length === 1 && firstLine.startsWith("/")` in render. The keypress handler computes `matches` on the full string, but `filterCommands` naturally returns `[]` once the input contains characters that no command name starts with (e.g. `\` or `\n`), so there's no latent popup-mutation bug to worry about. Verified by tracing the typed-`/help\`+Enter+more flow.
- Backspace-at-column-0 join is handled by the existing `deleteBeforeCursor` (it just deletes the inter-line `\n`) — nice that this required no new code.
- `done()` mirrors the multi-line shape so scrollback shows the user's final command as they entered it.

### One real concern — multi-line history persistence

[src/cli/input.ts:37-59](https://github.com/zjshen14/opencli/blob/feat/multiline-input-backslash-continuation/src/cli/input.ts#L37-L59) uses `\n` as the inter-entry separator in the on-disk history file:

```ts
// saveHistory
const lines = [...history].reverse().slice(-MAX_HISTORY).join("\n") + "\n";

// loadHistory
return raw.split("\n").map((l) => l.trim()).filter(Boolean).reverse();
```

Before this PR, every in-memory history entry was a single line, so the format worked. With this PR, an entry like `"Read this file:\n./src/foo.ts"` will be written verbatim, and on the next session `loadHistory` will split it back into two distinct entries (`"Read this file:"` and `"./src/foo.ts"`). Recalling history with Up arrow will then show fragmented junk where a multi-line prompt should be.

In-session history (within one process) works fine since the `history[]` array holds the full string. The regression is across-session only — but it's a real one that's hard to discover without trying.

**Suggested fix options** (in roughly increasing cost/correctness):
1. Skip persisting any entry that contains `\n` (`.filter((e) => !e.includes("\n"))` before `join`). Simplest, but you lose recall of expensive-to-retype multi-line prompts.
2. Escape `\n` → `\\n` on save, reverse on load. Tiny change.
3. Switch the on-disk format to one entry per JSONL line: each line is a JSON-encoded string. Aligns with the existing session storage pattern; future-proof against other separators.

Option 2 is probably the right pragmatic choice for this PR.

### Minor observations (non-blocking)

- The `\`+Enter handler resets `cursorPos = input.length` regardless of where the cursor was when Enter was pressed. So if a user typed `foo\` and then moved the cursor to the start before Enter, the cursor jumps to end-of-input after the continuation. Natural-feeling for the common case but worth a sentence in the comment if intentional.
- `cursorRowsBelowTop` math assumes no terminal-width line wrapping. Same limitation as the pre-PR renderer — not a regression, but worth tracking as a known limit on very long single lines.
- The MVP punt on in-buffer Up/Down navigation is well-flagged in the PR notes; agree it's a reasonable scope cut.

### Test coverage
- `cursorLineCol` is comprehensively unit-tested ✓.
- Render integration is acknowledged as untested due to no raw-mode fixture. Acceptable given the helper is pure and the integration is straight-line composition.
- No test asserts the new continuation behavior end-to-end (input `foo\` + Enter produces `foo\n` and re-renders) — would be worth adding a small test that drives `readLine` via a mock stdin if/when a raw-mode fixture lands, but not blocking now.

### Verdict
Not approving as-is — the history persistence regression should be addressed (option 2 above is 4–5 lines). Otherwise the PR is solid and merge-ready.

Issue #113, opened 2026-05-17
A5b: Auto-compact with safety gates (post-A5a validation)


zjshen14:
Auto-compact with opt-in config gate, token API integration, and failure handling. Blocked on A5a real-session Phase 2 inspection.

This is the second phase of context compaction. A5a ships manual `/compact` and `/context`; A5b adds automatic compaction after real-session data validates the approach.

## What A5b adds

- Auto-compact trigger at 75% of context window (token-based)
- Token API integration (Anthropic countTokens, Gemini, OpenAI tiktoken) for accurate estimates
- Original prompt preservation: keep first 2 messages verbatim
- Compaction failure handling: retry + fallback logic
- Turn-boundary detection: only trigger after agent completes full turn (avoids mid-task compaction)
- Threshold warnings: hint at 60%, stronger at 75%
- Config gate: `autoCompact: false` default; flip to `true` after validation
- Observability: metrics for frequency, tokens freed, cost/latency

## Blockers

1. A5a must ship and collect Phase 2 real-session data (1–2 weeks)
2. Manual inspection of real compacted sessions to validate quality
3. Design refinements based on Phase 2 findings

## Files touched

Enhance from A5a:
- `src/core/compact.ts` — add token API client, add `isAutoCompactNeeded()`
- `src/core/agent.ts` — add `compact` event, add auto-compact hook in main loop
- `src/cli/repl.ts` — add threshold warnings, handle `compact` events
- Config system — add `autoCompact: boolean` field

## Next: D1 integration

Extend D1 eval harness with long scenarios that trigger compaction, compare solve rates before/after auto-compact to validate correctness.

See [A5 design doc](https://github.com/zjshen14/opencli/blob/main/docs/design/a5-compact-context.md) for full spec.

Blocked by [#112 (A5a)](https://github.com/zjshen14/opencli/issues/112).

zjshen14:
Reviewed all 21 open issues during an autonomous agent run (2026-05-22). No new feasible PR was opened this run.

All small/well-scoped issues already have open PRs awaiting review:
- #119 → PR #129
- #116 → PR #130
- #128 → PR #131
- #117 → PR #132
- #118 → PR #133
- #125 → PR #135
- #127 → PR #136

[...]

PR #189, opened 2026-05-30
docs(A5b): promote A5b to ready-for-implementation with Phase 2 design decisions

zjshen14:
## Summary

The A5 design doc had A5b queued behind a Phase 2 data-collection step that's now done: the card_trade session \`2026-05-17T22-00-06-184\` ran across 662 events / 42 user turns / ~402k estimated tokens and gave us concrete numbers to resolve the deferred risks against.

This PR folds those Phase 2 findings into [docs/design/a5-compact-context.md](docs/design/a5-compact-context.md) and turns every \"TBD\" row in the A5b scope table into a concrete spec, with a file-level implementation outline and acceptance criteria.

## Key design decisions resolved

| Deferred risk | Resolution |
|---|---|
| **Where to fire** | Top of \`Agent.run()\`, after \`addMessage\` and before the stream loop — the one position where trajectory elongation is structurally impossible (no in-flight tool calls or LLM stream to interrupt). |
| **When to fire** | 75% of \`contextWindowFor(model)\`, with a 60% soft warning that fires once per session. |
| **Token accuracy** | Keep \`JSON.stringify.length / 4\`. Phase 2 confirms the constant-factor error absorbs into the threshold; provider token APIs add a round-trip for zero behavior gain. Estimator contract is isolated so it can be swapped later. |
| **Original prompt** | Moves into the summary message body (verbatim quotation at the top). PR #154's prune anchor stays as a safety net for short sessions. |
| **Failure mode** | Fail-open: catch, emit \`compact_failed\` observability, print stderr warning, continue with un-compacted history. Three reasons documented. |
| **Default** | \`autoCompact: true\` by default once implemented. Opt-out via \`~/.opencli/config.json\`. |
| **Observability** | Four new event types (\`compact_threshold_warned\`, \`compact_started\`, \`compact_completed\`, \`compact_failed\`) — unified surface for manual + auto paths. |

Plus a **replay test plan** using the card_trade session as a fixture, so the design is validated against the data it was tuned for.

## Phase 2 findings (the data the design rests on)

| Metric | Value | Implication |
|---|---|---|
| Total events | 662 | Real sessions exceed default \`historySize: 50\` by 10× |
| User turns | 42 | Manual \`/compact\` would need 21 invocations — unrealistic |
| Tool call/result pairs | 290 + 289 | Tool calls dominate context, matches JetBrains finding (~84% of tokens) |
| Estimated tokens | ~402k | Under Gemini's 1M window; over 100k fallback. Window-aware trigger is right. |
| Tool results with \`Error:\` | 23 | Verbatim error preservation matters in practice |
| First prune trigger | event #92 (~14% through) | Without auto-compact, the head is being dropped for 86% of the session |

## Related issue

Sets up the implementation track for issue [#113](https://github.com/zjshen14/opencli/issues/113) (A5b).

## Test plan

- [x] \`npm run typecheck && npm run lint && npm run format:check && npm test\` — pass (doc-only change)
- [x] Section structure reviewed — 7 new design decisions, each with rationale
- [x] Implementation outline maps each decision to a specific file change
- [ ] Implementation PR follows once this design lands

## Notes for reviewers

Status line moves A5b from \"blocked pending real-session data\" to \"ready for implementation\". The implementation should be a follow-up PR; this one is design-only so review can focus on whether each decision actually answers the deferred risk before we write code.

Two cross-references worth checking:
- The \`maybeAutoCompact()\` shape in section §5 (failure handling) is the proposed call signature for the implementation PR — feedback now is cheap.
- The verbatim-original-task quotation in §4 *replaces* the prune anchor as the canonical mechanism over time, but #154's anchor stays in place. Both paths coexist deliberately for the \`autoCompact: false\` case.

### DIFF
--- a/docs/design/a5-compact-context.md
+++ b/docs/design/a5-compact-context.md
@@ -155,7 +155,7 @@ The card_trade session (`2026-05-17T22-00-06-184`) ran across two real coding se
 | Total events in JSONL | 662 | Real sessions exceed the default `historySize: 50` by an order of magnitude |
 | User turns | 42 | Manual `/compact` would require 21 invocations to keep the head visible — unrealistic |
 | `tool_call` + `tool_result` pairs | 290 + 289 | Tool calls dominate context, exactly as the JetBrains paper predicted (~84% of tokens) |
-| Estimated tokens (JSON/4) | ~402k | Comfortably under Gemini's 1M window; well over the 100k unknown-model fallback. Threshold-based trigger needs the per-model window we already have. |
+| Estimated tokens (JSON/4) | ~402k | Under Gemini's raw 1M window (40%) but well over a sensibly-capped effective window — motivates the 256k cap in §2 |
 | Tool results containing `Error:` | 23 | Error signals are real and frequent — the verbatim-preservation rule from A5a matters |
 | First prune (50-message threshold) | event #92 | About 14% through; from there on the head was being dropped every turn |
 | `JSON.stringify.length / 4` vs. real token count | within ~2× for Gemini; doesn't matter for trigger | A constant-factor error absorbs into the threshold. Save the provider-API round-trip cost. |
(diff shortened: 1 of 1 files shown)


zjshen14:
## Response to review — all six points addressed in 4456465

Review was right on every count. Verified the threshold math by hand against the Phase 2 numbers — the original doc *did* argue against itself, and the fix is a 256k cap on the effective compaction window.

### Per-item fixes

**#1 (High) Threshold ↔ Phase 2 contradiction.**
Introduced \`COMPACTION_TARGET_TOKENS = 256_000\`. Trigger is now \`tokens / min(contextWindow, 256_000) >= 0.75\`. Math now adds up: at ~9.5k tokens/turn the card_trade session compacts at turn ~21 and ~32 = **exactly 2 compactions**, matching the original claim instead of contradicting it. Added a §2 table showing per-provider effective windows + rationale (the cap is also justified on Gemini cost-and-latency grounds).

**#2 (High) Prune ↔ compaction interaction.**
Added new §6 with the explicit invariants:
- Prune fires on every \`addMessage()\`; compaction fires at turn boundary.
- After compaction, history is 11 messages — below the 50-message prune cap.
- The summary message at position 0 is a user-role text message, so the PR #154 anchor logic preserves it through prune.
- Nested compactions preserve the original task via a verbatim-quotation rule in the structured prompt, by string equality not paraphrase.

**#3 (Medium) Call signature didn't match Agent's shape.**
Fixed to \`private autoCompact: boolean\` field (not \`this.config.autoCompact\`). The token estimate in \`maybeAutoCompact\` now includes \`systemInstruction\` so the trigger matches the actual \`client.stream()\` payload — added to §5 *and* to the acceptance criteria so the implementation can't drift.

**#4 (Medium) \`process.stderr.write\` from core/ breaks the established pattern.**
Replaced with a new \`AgentEvent\` variant: \`{ type: "notice"; message: string }\`. \`maybeAutoCompact()\` now returns \`Promise<AgentEvent[]>\` and \`run()\` yields them — mirrors the snapshot-warning pattern at \`agent.ts:249-252\`. Implementation outline adds the renderer change (notices render as a dim one-liner, not through the markdown stream).

**#5 (Medium) Replay test exercises zero compactions.**
Parametrize the model in the fixture — use \`claude-sonnet-4-6\` (200k window, trigger at ~150k tokens) so the session's ~9.5k-tokens/turn growth actually crosses the threshold. Added lower-bound assertion (≥ 1 compaction) alongside the existing ceiling (≤ 5). Test now has teeth.

**#6 (Low) 60% warning doesn't re-arm.**
Reset \`warnedAt60 = false\` on successful compaction. Documented in §2 and visible in the pseudocode in §5.

### Doc shape

Sections renumber cleanly 1-8 (added §6 for prune interaction; old §6 → §7, old §7 → §8). \`npm run format:check\` clean.

Re scope note in the review: confirmed — this PR is the auto-context-compaction A5b, not the subagent A5b. Both live in \`a5-compact-context.md\` as intended.

Re harness note: thanks for flagging. I've been editing through Edit/Read since the rebuild and haven't seen the empty-output issue, but I'll restart if it shows up.

Ready for re-review.

PR #191, opened 2026-05-30
feat(A5b): auto-compact at turn boundary with safety gates

### LINKED ISSUE #113: A5b: Auto-compact with safety gates (post-A5a validation)

zjshen14:
## Summary

Implements A5b per the design in [docs/design/a5-compact-context.md](docs/design/a5-compact-context.md) (PR #189). Auto-compact fires only at the top of \`Agent.run()\` — the one position where trajectory elongation is structurally impossible because no LLM stream or tool call is in flight.

## What's new

| Surface | Change |
|---|---|
| \`AgentEvent\` | New variant \`{ type: "notice"; message: string }\` for non-fatal informational signals; distinct from \`error\` so the renderer can style differently |
| \`Agent\` constructor | Optional \`autoCompact?: boolean\` (default \`true\`) |
| \`ObservabilityEvent\` | Four new types: \`compact_threshold_warned\`, \`compact_started\`, \`compact_completed\`, \`compact_failed\` — unified surface for manual and auto paths |
| \`Config\` | Optional \`autoCompact?: boolean\`. Default resolution \`autoCompact !== false\` |
| \`compact.ts\` | Exports \`COMPACTION_TARGET_TOKENS = 256_000\`. \`compactHistory()\` extracts and re-quotes the verbatim original task. Summarization prompt instructs verbatim copy of the quotation block for nested compactions |
| \`runner.ts\` | Renders \`notice\` events via \`printInfo\` |

## Behavior gates

```
ratio < 0.60         → no action
0.60 ≤ ratio < 0.75  → one notice per session (warnedAt60 latched);
                       re-armed by clearHistory() or successful compaction
ratio ≥ 0.75         → call compactHistory(); fail-open on error
                       (notice + compact_failed observability; turn proceeds)
```

The effective window is \`min(contextWindowFor(model), 256_000)\`. Without that cap a 75%-of-1M Gemini trigger would never fire in practice — the motivating card_trade session peaked at ~402k tokens (~40% of Gemini's raw window), which is why the design carries the cap.

Token estimate intentionally includes \`systemInstruction\` so the threshold reflects the real payload sent to \`client.stream()\`, not just the message bytes.

## Related issue

Closes #113

## Test plan

- [x] \`npm run typecheck && npm run lint && npm run format:check && npm test\` — 643 tests pass (+13 new)
- [x] **10 new agent-level tests** (\`agent.test.ts\`): no-fire below 60%, warning between 60-75%, warning once-per-session, auto-fire at 75%, \`autoCompact: false\` disables, **256K-cap test confirms Gemini fires at 200K tokens** (would never fire without the cap), warning re-arms after compaction, fail-open yields notice without throwing, \`clearHistory()\` resets the warning flag
- [x] **3 new compact tests** (\`compact.test.ts\`): verbatim quotation prepended, original task survives nested compaction by string equality (not paraphrase), no-anchor branch when head has no user-text
- [ ] Manual: run a long REPL session past 60% and 75% to observe the notices and the compaction in real terminal output

## Acceptance criteria (from the design doc, all satisfied)

| Criterion | Status |
|---|---|
| Fires only at turn boundary | ✅ at top of \`run()\` after \`addMessage\`, before stream loop |
| Uses \`min(window, 256_000)\` for effective trigger | ✅ verified by the Gemini-200K test |
| \`autoCompact: false\` fully disables | ✅ both auto-fire and 60% warning suppressed |
| Warning fires once per session, re-arms after compaction | ✅ \`warnedAt60\` reset on successful compaction |
| Failures don't block the turn | ✅ \`fail-open\` test confirms |
| Original task survives via verbatim quotation, including nested | ✅ string-equality test confirms |
| Token estimate includes \`systemInstruction\` | ✅ in \`maybeAutoCompact\` |
| Notices via AgentEvent stream (not \`process.stderr\`) | ✅ \`notice\` event renders through \`printInfo\` |

## Notes for reviewers

The implementation closely tracks the design doc — every numbered section in §1-§8 has a matching code change. Worth checking the two judgment calls:

1. **Verbatim quotation extraction is permissive**: \`extractOriginalTask()\` searches the entire head for the first user-role text message rather than only position 0. Reason: by the time the second compaction runs, the head includes a previous summary (which is itself a user-text message starting with \`[Session context compacted ...\`); the function correctly looks past it to find the embedded \`**Original task** (verbatim):\` block and copies its inner lines forward. That's the nested-preservation invariant in §6.

2. **Manual \`/compact\` observability**: I emit \`compact_started\`/\`compact_completed\` from \`Agent.compact()\` rather than from the \`/compact\` REPL handler. Reason: \`/compact\` is one of several ways to trigger manual compaction (e.g. tests, programmatic use); emitting from \`Agent.compact()\` covers them all. Manual path errors still propagate (unlike auto-fail-open) since the user explicitly asked.

### DIFF
--- a/src/core/agent.test.ts
+++ b/src/core/agent.test.ts
@@ -620,8 +620,11 @@ describe("Agent auto-compact (A5b)", () => {
     const notice = yielded.find((e) => e.type === "notice");
     expect(notice).toBeDefined();
     expect((notice as { type: "notice"; message: string }).message).toMatch(
-      /auto-compact will trigger at 75%/,
+      /approaching auto-compact threshold/,
     );
+    // Shows tokens/budget pair so the user doesn't read it as a fraction of
+    // the model's full context window.
+    expect((notice as { type: "notice"; message: string }).message).toMatch(/\d+k \/ \d+k tokens/);
   });
 
   it("fires the warning only once per session across consecutive turns", async () => {
@@ -664,25 +667,53 @@ describe("Agent auto-compact (A5b)", () => {
   });
 
   it("re-arms the 60% warning after a successful compaction", async () => {
-    // Start above 75% so the first turn auto-compacts (and re-arms the
-    // warning). Then force the ratio back to 65% and confirm a second
-    // warning fires.
-    const { agent, events } = makeAgentAtRatio(0.8, "claude-sonnet-4-6");
+    // Genuine re-arm coverage: the previous version went straight to 0.8 and
+    // skipped the 60–75% latch entirely, so deleting the post-compaction
+    // re-arm line would still have passed it. We now:
+    //   1. Sit at 0.65 → first warning fires; warnedAt60 latches true
+    //   2. Climb to 0.80 → auto-compact fires; re-arm code resets to false
+    //   3. Drop to 0.65 again → SECOND warning must fire (proves re-arm)
+    const { agent, events } = makeAgentAtRatio(0.65, "claude-sonnet-4-6");
     await collectEvents(agent, "first");
-    expect(events.filter((e) => e.type === "compact_completed")).toHaveLength(1);
-    // After compaction the history shrank — re-inflate to the 65% band.
-    // Easier: directly re-set messages to a known-size payload.
-    const padBytes = Math.ceil(200_000 * 0.65) * 4 - 200;
-    const refilled = [
-      { role: "user" as const, parts: [{ type: "text" as const, text: "ORIGINAL_TASK_TOKEN" }] },
-      { role: "model" as const, parts: [{ type: "text" as const, text: "x".repeat(padBytes) }] },
-    ];
-    agent.restoreMessages(refilled);
-
-    await collectEvents(agent, "second");
-    // Two warnings total: one initial-trigger path is N/A because the first
-    // turn hit 80%, but the new climb into 65% must rearm and warn.
     expect(events.filter((e) => e.type === "compact_threshold_warned")).toHaveLength(1);
+    expect(events.filter((e) => e.type === "compact_completed")).toHaveLength(0);
+
+    // Climb into the auto-compact band.
+    const effectiveWindow = Math.min(
+      contextWindowFor("claude-sonnet-4-6"),
+      COMPACTION_TARGET_TOKENS,
+    );
+    const fill = (ratio: number) => {
+      const padBytes = Math.ceil(effectiveWindow * ratio) * 4 - 200;
+      agent.restoreMessages([
+        { role: "user", parts: [{ type: "text", text: "ORIGINAL_TASK_TOKEN" }] },
+        { role: "model", parts: [{ type: "text", text: "x".repeat(padBytes) }] },
+      ]);
+    };
+
+    fill(0.8);
+    await collectEvents(agent, "second");
+    expect(events.filter((e) => e.type === "compact_completed")).toHaveLength(1);
+
+    // Re-fill the 60–75% band. If re-arm is broken, no second warning.
+    fill(0.65);
+    await collectEvents(agent, "third");
+
+    expect(events.filter((e) => e.type === "compact_threshold_warned")).toHaveLength(2);
+  });
+
+  it("does not run auto-compact in plan mode (read-only exploration)", async () => {
+    const { agent, events } = makeAgentAtRatio(0.8, "claude-sonnet-4-6");
+
+    // Drain a plan-mode turn. Even though ratio is above the trigger,
+    // plan mode is read-only exploration and shouldn't spend tokens on a
+    // compaction round-trip — the next react turn will trigger it.
+    const planEvents = [];
+    for await (const e of agent.run("plan something", "plan")) planEvents.push(e);
+
+    expect(events.some((e) => e.type === "compact_started")).toBe(false);
+    expect(events.some((e) => e.type === "compact_completed")).toBe(false);
+    expect(planEvents.some((e) => e.type === "notice")).toBe(false);
   });
 
   it("fail-open: compactionClient error yields a notice and does not throw", async () => {
--- a/src/core/agent.ts
+++ b/src/core/agent.ts
@@ -135,8 +135,12 @@ export class Agent {
     // A5b auto-compact: fires only at this turn-boundary position — before any
     // LLM call or tool execution this turn. Any LLM/network errors inside
     // maybeAutoCompact are caught by it; the turn always proceeds.
-    for (const notice of await this.maybeAutoCompact(systemInstruction)) {
-      yield notice;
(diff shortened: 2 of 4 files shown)


zjshen14:
## Review — A5b auto-compact implementation

Solid PR, and it cleanly addresses the feedback from the #189 design review: notices flow through the `AgentEvent` stream rather than `process.stderr`, the token estimate now includes `systemInstruction`, the `min(window, 256_000)` cap fixes the "75%-of-1M Gemini never fires" problem, the 60% warning re-arms, and compaction is fail-open. The structure tracks the design section-by-section. A few things worth addressing before merge.

### 1. (Medium) Original task is duplicated on nested compactions
`compactHistory()` extracts the verbatim task and **prepends** `originalTaskBlock` to the summary message itself, *and* `SUMMARIZATION_PROMPT` now instructs the LLM to "copy that block VERBATIM as the first thing in your response." On a nested compaction the head contains the previous summary (which holds the block), so the LLM copies it into `summary` **and** the code prepends it again — the `**Original task** (verbatim):` block ends up in the output twice.

Since `extractOriginalTask()` already reconstructs the block deterministically in code, the prompt rule is redundant and is the source of the duplication. Suggest letting the code own preservation and dropping (or inverting) the prompt instruction so the model does *not* re-emit the block. The nested test passes today only because it asserts `toContain(...)`, which a double-inclusion also satisfies.

### 2. (Medium) The "re-arms after compaction" test doesn't actually exercise re-arm
In `re-arms the 60% warning after a successful compaction`, the agent starts at ratio 0.8, so the first turn goes straight to the compact branch and **never latches `warnedAt60 = true`** (the 60–75% branch is skipped). After compaction `warnedAt60` is set back to `false` — but it was already `false`. The second turn at 0.65 then produces the first-ever warning. So the test would pass even if the post-compaction re-arm line were deleted.

To genuinely cover re-arm: warn at 0.65 (latch `true`) → climb to ≥0.75 (compact, re-arm) → drop to 0.65 again → assert a **second** `compact_threshold_warned` (length 2).

### 3. (Low–Med) Displayed percentage is relative to the capped window, not the model's real window
`ratio` is computed against `min(contextWindowFor(model), 256_000)`, and the notice prints `context at ${Math.round(ratio*100)}%`. On Gemini (1M window) a user sitting at ~154K tokens is at ~15% of their real window but will see "context at 60%." That's arguably confusing. Consider wording it relative to the compaction threshold (e.g. "approaching the auto-compact threshold") or naming the effective budget, so the number doesn't read as "60% of my model's context."

### 4. (Low) Confirm the prune-anchor interaction with `extractOriginalTask`
`extractOriginalTask(head)` returns the first user-role text message. This is correct only as long as the original first user message is still at the head of context when compaction first fires. In a long session, message-count pruning runs well before the ~192K-token trigger. This should be safe because the #154 prune anchor keeps the first user message — but that's the load-bearing assumption for the whole "original task survives" feature, and there's no test pinning it. Worth an explicit test: prune past the cap, then auto-compact, and assert the *real* original task (not a later user turn) is what gets quoted.

### 5. (Low) Auto-compact also runs in plan mode
`maybeAutoCompact()` runs at the top of `run()` regardless of `mode`, so a read-only `/plan` turn can trigger a compaction-client round-trip. Probably harmless, but compacting during read-only exploration is a little surprising and spends tokens — consider skipping when `mode === "plan"`.

### Nits
- `maybeAutoCompact` recomputes the token estimate that the turn loop also computes a few lines later for `context_snapshot`. Minor; could share.
- `COMPACTION_TARGET_TOKENS = 256_000` means Gemini compacts at ~192K despite a 1M window — intended per design §2, just flagging the aggressiveness for the record.

Nothing here is blocking on its own; 1 and 2 are the two I'd want resolved (one is a real output artifact, the other a test that doesn't test what it claims).

zjshen14:
## Re-review — follow-up commit `501862e7` ✅

All five points are addressed, and point 4 went beyond the request — it fixed a real latent bug, not just added a test.

1. **Duplicate original-task block — fixed.** The `SUMMARIZATION_PROMPT` rule is now *inverted* (instructs the model to OMIT any existing block); the calling code owns preservation and prepends exactly once. The nested-compaction case can no longer double-include.

2. **`re-arms 60% warning` test — now genuine.** Rewritten to warn at 0.65 (latch) → compact at 0.80 (re-arm) → drop to 0.65 → assert a **second** `compact_threshold_warned`. Deleting the re-arm line would now fail it.

3. **Notice wording — fixed.** Now reads `approaching auto-compact threshold — Xk / Yk tokens used (compacts at 75% of budget)`, framed against the compaction budget rather than a misleading "% of model window."

4. **Prune × `extractOriginalTask` — fixed (caught a real bug).** PR #154's anchor merges the original task with the next user turn via `[earlier conversation pruned]`; `extractOriginalTask` was returning the whole merged string, leaking the later turn into the quote. Now strips at the marker, with a test driving the full long-history → prune → compact path.

5. **Plan mode — fixed.** `maybeAutoCompact()` is now guarded by `mode !== "plan"` so read-only exploration doesn't spend tokens on a compaction round-trip.

Verified locally on `501862e7`: `typecheck` + `lint` + `format:check` clean; `compact.test.ts` + `agent.test.ts` → **54 passed**.

LGTM — merging.

PR #197, opened 2026-05-31
fix(A5b): address five review points from PR #191

zjshen14:
## Summary

Re-opens the A5b review feedback as a clean PR off main. The original PR #191 became confusing after its feature commit landed on main through another path; this PR contains only the post-review fixes against the current main.

### 1. Original task no longer duplicated on nested compactions
\`SUMMARIZATION_PROMPT\` previously instructed the model to copy the \`**Original task** (verbatim):\` block *and* the calling code prepended its own copy — yielding the block twice in any nested compaction. Inverted the prompt rule to **OMIT** any existing block; the code owns preservation deterministically.

### 2. The "re-arms 60% warning" test now actually tests re-arm
Previous version started at 0.8, skipped the 60-75% latch, and would have passed even if the post-compaction reset line were deleted. Rewritten to:
1. Sit at 0.65 → latch \`warnedAt60 = true\`
2. Climb to 0.80 → compact, re-arm
3. Drop to 0.65 → assert a SECOND \`compact_threshold_warned\` (length 2)

### 3. Notice wording reframed to the compaction budget
Now reads \`approaching auto-compact threshold — 154k / 256k tokens used (compacts at 75% of budget)\` — making clear this is OpenCLI's threshold, not the model's window. Test asserts the new wording and the \`tokens used\` shape.

### 4. Prune × \`extractOriginalTask\` interaction tested and fixed
Surfaced a real bug. PR #154's prune anchor merges the original task with the next user turn via \`\\n\\n[earlier conversation pruned]\\n\\n\`. \`extractOriginalTask\` returned the merged text — which would have leaked the next user turn into the quotation.

Fix: \`extractOriginalTask\` now recognises three shapes (plain text, prune-merged, nested compaction) and strips the merge marker. New integration test exercises the full path (\`maxHistoryMessages = 25\` → addMessage in a loop until prune fires repeatedly → compact → assert the quotation contains only the original task).

### 5. Auto-compact skipped in plan mode
\`maybeAutoCompact()\` previously ran on every \`Agent.run()\` including plan-mode passes. Plan is read-only exploration; spending tokens on a compaction round-trip there is surprising and unnecessary. The react turn that follows handles it. New test asserts no \`compact_started\`/\`compact_completed\`/\`notice\` from a plan-mode turn even at ratio 0.8.

## Related issue / PR

Follow-up to the merged-by-other-means A5b feature in main. Original review on PR #191 (closed alongside this).

## Test plan

- [x] \`npm run typecheck && npm run lint && npm run format:check && npm test\` — 645 tests pass (+3 net new)
- [x] All five review items have a corresponding new or rewritten test
- [ ] Manual: run a session past 60% with the new wording visible

## Notes for reviewers

The two unfixed nits called out in the review (duplicate token-estimate computation in \`maybeAutoCompact\` and the aggressive 256K threshold for Gemini) are intentional — flagged in the review reply on PR #191.

### DIFF
--- a/src/core/agent.test.ts
+++ b/src/core/agent.test.ts
@@ -620,8 +620,11 @@ describe("Agent auto-compact (A5b)", () => {
     const notice = yielded.find((e) => e.type === "notice");
     expect(notice).toBeDefined();
     expect((notice as { type: "notice"; message: string }).message).toMatch(
-      /auto-compact will trigger at 75%/,
+      /approaching auto-compact threshold/,
     );
+    // Shows tokens/budget pair so the user doesn't read it as a fraction of
+    // the model's full context window.
+    expect((notice as { type: "notice"; message: string }).message).toMatch(/\d+k \/ \d+k tokens/);
   });
 
   it("fires the warning only once per session across consecutive turns", async () => {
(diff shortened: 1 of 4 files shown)


zjshen14:
Closing as superseded. The same five review fixes for A5b are already on main via commit 0042ada (the original PR #191 was re-opened and merged with the fixes included), so this PR's diff is empty against current main. Thanks anyway.

Issue #60, opened 2026-05-03
Architecture: replace agentic while-loop with an explicit state machine


zjshen14:
## Background

`Agent.run()` in `src/agent/core.ts` is a single `while(true)` loop managing multiple mutable variables — `turns`, `lastCallSig`, `stuckCount`, `pendingCalls`, `responseText` — all in one function. The phases of the loop (streaming, collecting tool calls, executing, feeding results back) are implicit, interleaved in the same control flow.

## Problems this causes

- **Safety guards are hardcoded**: max-turns and stuck-loop detection are if-statements in the loop body. Adding a new guard (cost limit, time limit, human interrupt checkpoint) means editing the loop itself.
- **No middleware layer**: there is no clean injection point for cross-cutting concerns (observability, rate limiting, token tracking) between phases.
- **Hard to test phase transitions**: to unit-test stuck-loop detection you must run the full loop; you can't test the guard in isolation.
- **Pause/resume is impossible**: if the user presses Ctrl+C mid-execution, there is no clean way to capture and restore the in-flight state.

## Proposed design

Model the loop as an explicit state machine with named states and named transitions:

```
IDLE
  → STREAMING        (LLM generating text + function calls)
  → EXECUTING_TOOLS  (parallel/sequential tool dispatch)
  → FEEDING_RESULTS  (tool results assembled into user message)
  → DONE
  → ERROR
```

Each state is a method; each transition is a function call. Guards (max-turns, stuck-loop, cost limit) become composable `TransitionGuard` objects that run at well-defined checkpoints, not scattered if-statements.

```ts
interface AgentState { type: "idle" | "streaming" | "executing" | "feeding" | "done" | "error" }
interface TransitionGuard {
  check(state: AgentState, context: AgentContext): GuardResult; // pass | block(reason)
}
```

This is the approach taken by production agent frameworks (LangGraph, Temporal workflows) — the state graph *is* the architecture.

## Acceptance criteria

- [ ] Agentic loop phases are represented as distinct, named states
- [ ] Max-turns and stuck-loop detection are `TransitionGuard` implementations, not inline if-statements
- [ ] A new guard can be added by implementing `TransitionGuard` and registering it — no changes to the core loop
- [ ] Existing behaviour (streaming, tool execution, plan mode) is preserved
- [ ] Unit tests cover individual guards in isolation

## Location

`src/agent/core.ts` — `Agent.run()`

PR #199, opened 2026-05-31
refactor(agent): extract run() loop body into focused helper methods

zjshen14:
## Summary

\`run()\` had grown to 224 lines doing five distinct jobs interleaved: turn setup, stream collection, exit checks, four safety guards (max-turns, stuck-loop, env-error, empty-retry), tool execution, snapshot warning drain, reminder synthesis, and event yielding. Every new guard since A1 made the diff uglier and the tests harder to write.

This PR splits the loop body into named private methods, leaving \`run()\` as a clean 10-step pipeline:

\`\`\`
1. streamLLM(setup)                       → text + pendingCalls + obs events
2. recordAssistantMessage(...)            → context mutation
3. exit checks (done / empty-retry)       → inline
4. checkMaxTurnsGuard / checkStuckLoopGuard → abort events
5. executeTurnTools(...)                  → tool execution
6. drainSnapshotWarnings()                → error event if any
7. checkEnvErrorGuard(...)                → abort event if hit
8. applyReminderToLastResult(...)         → mutates results in place
9. yieldSkillActivations + yieldToolResults → events to caller
10. recordToolResults(...)                → context mutation
\`\`\`

## Numbers

| | Before | After |
|---|---|---|
| \`run()\` body | 224 lines | 88 lines (**-60%**) |
| File total | 455 lines | 558 lines (+23%, mostly method signatures + JSDoc) |
| Tests | 645 passing | **645 passing — no test modifications** |
| Public API | unchanged | unchanged |

## Notable shape decisions

- \`streamLLM\` is \`AsyncGenerator<AgentEvent, { text, pendingCalls }>\` — uses TS's generator return-value typing so the caller can \`yield*\` for the events and destructure the collected state in one statement.
- Guard methods return \`AgentEvent | null\` rather than throwing; the caller yields-and-returns. Mirrors the snapshot-warning yield pattern already in the loop. Easy to test (call the method on a \`TurnState\` fixture, assert the returned event).
- \`TurnState\` is a plain interface (not a class) so guards can mutate counters without ceremony. Pure value type, easy to construct in tests.

## Related issue

Tracks [#60 — replace agentic while-loop with explicit state machine](https://github.com/zjshen14/opencli/issues/60). The extracted methods are exactly the transitions the state machine would exhibit; this PR lands the readability win now without committing to a specific state-machine library or shape. The full state machine refactor (formal \`State\` enum, transition table, etc.) can layer on top if/when we need it.

## Test plan

- [x] \`npm run typecheck && npm run lint && npm run format:check && npm test\` — 645 tests pass
- [x] No test file modifications (full coverage of behavior carried forward)
- [x] \`run()\` body now reads as a numbered sequence; each step delegates to a named helper
- [ ] Manual: run a real REPL session, confirm tool calls, snapshots, env-error guard, and auto-compact all work as before

## Notes for reviewers

This is a pure refactor — no behavior change. The right way to review is to skim each extracted method and verify it captures the exact lines it replaced. The four most subtle ones:

1. **\`streamLLM\`** — captures the LLM stream collection plus the \`context_snapshot\` / \`llm_call_start\` / \`llm_call_end\` observability triplet. Uses generator return-value typing.
2. **\`checkStuckLoopGuard\`** — the \"different signature, reset to 1\" branch is now an early return rather than an else-block, but the behavior is the same. \`stuckCount = 1\` (not 0) on first sight matches the previous code.
3. **\`checkEnvErrorGuard\`** — same: \"no match, reset counters\" branch is an early return; the rest matches.
4. **\`drainSnapshotWarnings\`** — sync generator (not async) because \`snapshotManager.drainWarning()\` is sync; both work as \`yield*\` targets.

The maybeAutoCompact() gate, manual \`compact()\` method, \`getContextStats()\`, \`undoLastTurn()\`, and \`clearHistory()\` are untouched.

### DIFF
--- a/src/core/agent.ts
+++ b/src/core/agent.ts
@@ -6,7 +6,12 @@ import { ContextManager } from "./context.js";
 import { executeCalls } from "./executor.js";
 import type { ConfirmFn } from "./executor.js";
 import { buildReminder, buildPlanSuffix } from "./prompt.js";
-import type { FunctionCallPart, Message } from "../providers/types.js";
+import type {
+  FunctionCallPart,
+  FunctionResultPart,
+  Message,
+  ToolDefinition,
+} from "../providers/types.js";
 import type { ObservabilityHandler } from "./observability.js";
 import type { SnapshotManager } from "../state/snapshot.js";
 import { compactHistory, contextWindowFor, COMPACTION_TARGET_TOKENS } from "./compact.js";
@@ -34,6 +39,25 @@ const AUTO_COMPACT_TRIGGER_RATIO = 0.75;
 const DEFAULT_MAX_TURNS = 50;
 const STUCK_THRESHOLD = 3;
 const ENV_ERROR_THRESHOLD = 3;
+
+// Per-turn mutable state passed through the run() pipeline. Each guard reads
+// and updates the fields it owns; the loop is otherwise stateless. Kept as a
+// plain interface (not a class) so guard methods can mutate counters without
+// ceremony.
+interface TurnState {
+  turns: number;
+  lastCallSig: string;
+  stuckCount: number;
+  envErrorPattern: string;
+  envErrorCount: number;
+  emptyRetried: boolean;
+  firedReminders: Set<string>;
+}
+
+interface TurnSetup {
+  toolDefinitions: ToolDefinition[];
+  systemInstruction: string;
+}
 
 // OS-level errors that code changes cannot fix. If the same pattern appears in
 // tool results across ENV_ERROR_THRESHOLD consecutive turns, the loop aborts.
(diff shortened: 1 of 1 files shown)


zjshen14:
## Review — refactor: extract `run()` loop body into helpers

Clean, well-motivated refactor. I read each extracted method against the lines it replaces and ran the suite locally — this is a faithful, behavior-preserving decomposition.

- `typecheck` + `lint` + `format:check` clean
- `src/core/` → **169 tests pass**, zero test-file changes (the right way to prove a pure refactor)

**Equivalence checks I paid attention to (all hold):**
- `streamLLM` — the `AsyncGenerator<AgentEvent, {text, pendingCalls}>` return-typing is a nice touch; the `context_snapshot` / `llm_call_start` / `llm_call_end` observability triplet and `usage` accumulation are preserved, and `llm_call_end` still fires after the stream drains.
- `checkStuckLoopGuard` — the "different signature → reset" path is now an early return with `stuckCount = 1` (not 0) on first sight, matching the original.
- `checkEnvErrorGuard` — counter reset on no-match, increment-vs-reset on pattern change, and the `>= ENV_ERROR_THRESHOLD` abort all match. Hoisting `.toLowerCase()` onto `combinedResults` once (instead of per-iteration) is a harmless micro-win with identical results.
- Step ordering is preserved end-to-end: stream → record assistant → done/empty-retry → max-turns → stuck → execute → **drain snapshot warning → env guard** → reminder → yield skill/results → record results. The snapshot-drain-before-env-guard order (which matters, since both can emit) is intact.
- `recordToolResults` early-returns on empty, same as the old `if (results.length > 0)` guard.

**Minor / non-blocking:**
- `yieldSkillActivations` now iterates all `pendingCalls` with an inner `if (name === "activate_skill")` instead of a pre-`filter`. Identical output; fine.
- Worth a quick manual REPL pass (the unchecked box in your test plan) just to confirm the env-error guard and auto-compact feel unchanged in a live session — but nothing in the diff suggests they would.

The `#60` framing is reasonable: these methods are exactly the transitions a future state machine would formalize, and landing the readability win without committing to a library is the right call.

LGTM ✅ — approving and merging.

PR #209, opened 2026-06-03
docs: fix architecture.md — add multi_edit to reminders and confirmation table

zjshen14:
## Summary

- The AGENT_REMINDERS description in `docs/architecture.md` said "write or edit" but the code (`isWriteCall` in `src/core/prompt.ts`) has included `multi_edit` since PR #208.
- The `requiresConfirmation` policy table was missing `multi_edit`, which triggers confirmation for paths outside `process.cwd()` — identical behavior to `edit`.

These two omissions were left after PR #207 (which added `multi_edit` to the file lists and `Tool` interface docs) and PR #208 (which fixed the code). This PR closes the remaining doc gap.

## Test plan
- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — all pass (642 tests, 0 failures)
- [x] No code changes — docs only; no runtime behavior is affected

Part of #43

https://claude.ai/code/session_01NRhVj7LV8swkrfMhJcVr5c

---
_Generated by [Claude Code](https://claude.ai/code/session_01NRhVj7LV8swkrfMhJcVr5c)_

PR #207, opened 2026-06-02
docs: add multi_edit and tool composition fields to source structure docs

zjshen14:
## Summary

- `multi_edit` was added in #193 but `CLAUDE.md`, `AGENTS.md`, and `docs/architecture.md` were never updated to reflect it
- Add `multi_edit` to the `file/` tool list in `CLAUDE.md` and `AGENTS.md`
- Add `multi-edit.ts` to the `file/` section of the `src/tools/` tree in `docs/architecture.md`
- Update the Tool interface code block in `docs/architecture.md` to include `ToolExecutionContext`, the updated `execute(params, ctx?)` signature, and the new `composedOf` / `singleConfirmation` composition fields

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — all pass (641 tests, 0 failures)
- [x] Doc-only change — no runtime behaviour changed; verified diff shows only the three doc files

https://claude.ai/code/session_01B4z4YjgZqJufbmQwp272qF

---
_Generated by [Claude Code](https://claude.ai/code/session_01B4z4YjgZqJufbmQwp272qF)_

PR #210, opened 2026-06-03
fix(renderer): handle multi_edit in toolStyle, formatToolArgs, and summariseResult

zjshen14:
## Summary

- `toolStyle("multi_edit")` was returning cyan/⟳ (the unknown-tool fallback) instead of yellow/✎ to match `edit` and `write`
- `formatToolArgs("multi_edit", { file_path, edits: [...] })` was falling through to the generic path-with-extras case, which JSON-dumps the `edits` array — potentially hundreds of lines of `old_string`/`new_string` content — into the bordered terminal box; now renders as `file.ts  (N edits)`
- `summariseResult("multi_edit", result)` fell through to the generic `name.padEnd(6)` case; since "multi_edit" is 10 chars, padEnd(6) is a no-op and the name was concatenated directly onto the result string without a space (e.g. `"multi_editApplied 3 edits…"`); now has an explicit case like `ls`, `write`, and `todo`

These were introduced when `multi_edit` was added but the renderer wasn't updated. Previous follow-up PRs (#207, #208, #209) fixed docs and `AGENT_REMINDERS` but this renderer gap remained.

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — all pass (648 tests)
- [x] 5 new tests added to `renderer.test.ts` covering the three fixed functions for `multi_edit`
- [x] Verified `toolStyle("multi_edit")` → `{ color: "yellow", icon: "✎" }`
- [x] Verified `formatToolArgs("multi_edit", { file_path: "src/foo.ts", edits: [{...}, {...}] })` → `"src/foo.ts  (2 edits)"` (no edits JSON leaked)
- [x] Verified `summariseResult("multi_edit", "Applied 3 edits to src/foo.ts")` → `"multi_edit  Applied 3 edits…"` (space preserved)

https://claude.ai/code/session_01GJvsuynvCqumP8f4SPMS3V

---
_Generated by [Claude Code](https://claude.ai/code/session_01GJvsuynvCqumP8f4SPMS3V)_

### DIFF
--- a/src/cli/renderer.test.ts
+++ b/src/cli/renderer.test.ts
@@ -56,6 +56,12 @@ describe("toolStyle", () => {
     expect(icon).toBe("✎");
   });
 
+  it("returns yellow + ✎ for multi_edit", () => {
+    const { color, icon } = toolStyle("multi_edit");
+    expect(color).toBe("yellow");
+    expect(icon).toBe("✎");
+  });
+
   it("returns cyan + ⟳ for unknown tools", () => {
     const { color, icon } = toolStyle("read");
     expect(color).toBe("cyan");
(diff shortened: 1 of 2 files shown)


zjshen14:
Verified: all three CI checks green, `mergeable_state` clean. The three renderer fixes are each correct and minimal — `toolStyle` delegates `multi_edit` to the same `yellow/✎` branch as `edit` and `write`; `formatToolArgs` renders `file.ts  (N edits)` without leaking `old_string`/`new_string` content; `summariseResult` applies a fixed two-space separator so the 10-char name doesn't collide with the result string. Five new tests cover each case including the singular/plural edge case and the no-JSON-leak invariant. Merging.



---

Issue #19, opened 2026-04-26
Design: ReAct loop quality improvements + dual-mode execution (ReAct vs Planner)


zjshen14:
## Background

Research into Claude Code, OpenHands, Cline, Aider, Goose, and Cursor reveals a consistent set of techniques the best agents use to improve loop quality. This issue proposes adopting the most impactful ones and adds a broader architectural question: should OpenCLI support both a **ReAct loop** and a **Planner mode**, and when should each be used?

---

### 1. Tool output truncation (middle-truncation)

**Current state:** Tool results pass through to context raw with no size limit. A single `bash` call outputting 10,000 lines consumes most of the context window.

**What others do:**
- Claude Code: 30k char hard cap, **middle-truncation** (head + tail preserved, middle elided), configurable via `BASH_MAX_OUTPUT_LENGTH`
- OpenHands: 30k chars, saves full output to a temp file, gives agent the path to read if needed
- Cursor: 20k chars (tail-only — users report this as a frustration)

**Proposal:**
- Default cap: **20k characters** per tool result
- Strategy: **middle-truncation** — preserve the first ~30% and last ~70% of output; the head contains command context, the tail contains the most recent errors
- For bash: append `[... N lines truncated — full output saved to {SESSION_TMP}/tool-output-<id>.txt ...]` so agent can read it if needed
- Configurable via `OPENCLI_MAX_TOOL_OUTPUT` env var

---

### 2. Max iterations guard

**Current state:** `while (true)` in `core.ts` — no ceiling. Issue #4 covers failing tools but not runaway successful loops.

**What others do:**
- Claude Code: `maxTurns` param (no default, but docs recommend 30 for production)
- OpenHands: 500 iterations default, `GLOBAL_MAX_ITERATIONS` for nested agents
- Goose: `--max-turns 1000`, separate `--max-tool-repetitions` for consecutive identical calls

**Proposal:**
- Add `maxTurns?: number` to `Agent` constructor (default: **50**)
- On breach: yield a final `{ type: "error", message: "Maximum iterations (N) reached. The task may be too complex — try breaking it into smaller steps." }` event and return
- Add `--max-turns` CLI flag to `opencli chat` / `opencli run`
- Separately: detect **stuck loops** (same tool + same args called 3× consecutively) and break early with a targeted message

---

### 3. Git context injected at session start

**Current state:** System prompt injects `{CWD}` but no git state. Agent wastes its first turn running `git status` / `git log` on almost every coding task.

[...]

PR #211, opened 2026-06-04
feat: inject periodic mid-loop reminder every 5 turns to counter instruction drift

zjshen14:
## Summary

- Adds `buildPeriodicReminder(turn)` to `src/core/prompt.ts` that returns a compact `[reminder: ...]` block every `PERIODIC_REMINDER_INTERVAL` (5) turns, and empty string otherwise
- Wires it into `applyReminderToLastResult` in `src/core/agent.ts` alongside the existing event-driven reminders
- Adds 3 unit tests in `prompt.test.ts` and 3 integration tests in `agent.test.ts` covering: fires at multiples of the interval, silent before the interval, fires again at the second multiple

## Related issue

Part of #19 — item 5: "Mid-loop system-reminder injection"

## Why

Event-driven reminders (`AGENT_REMINDERS`) fire at most once per run and only when specific tools are called (writes, git). In long sessions the model drifts from git rules, edit discipline, and testing requirements. A periodic reminder re-anchors it cheaply (~10 tokens every 5 turns) without disrupting short sessions where the interval is never reached.

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — all pass (654 tests, 0 failures)
- [x] New `buildPeriodicReminder` unit tests: fires at turn 5, 10; silent at turns 1–4; silent at turn 0
- [x] Agent integration tests: reminder appears in the `tool_result` event at exactly turn 5 and turn 10; absent before

🤖 Generated with [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01WwDBfCRj6VdggzUNJpudh9)_

### DIFF
--- a/src/core/agent.test.ts
+++ b/src/core/agent.test.ts
@@ -5,6 +5,7 @@ import { SkillRegistry } from "../skills/registry.js";
 import type { LLMClient } from "../providers/client.js";
 import type { Message, StreamEvent, ToolDefinition } from "../providers/types.js";
 import { contextWindowFor, COMPACTION_TARGET_TOKENS } from "./compact.js";
+import { PERIODIC_REMINDER_INTERVAL } from "./prompt.js";
 
 // A client that always requests the same tool call (never finishes on its own)
 function makeLoopingClient(toolName = "noop", args: Record<string, unknown> = {}): LLMClient {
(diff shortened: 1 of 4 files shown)


zjshen14:
## Review: feat: inject periodic mid-loop reminder every 5 turns

**What's good:** The implementation is clean and well-contained. `buildPeriodicReminder` is a pure function, correctly placed in `prompt.ts`, and wired into the existing `applyReminderToLastResult` path with minimal structural disruption. Tests are thorough: unit tests cover turns 0, 1–4, 5, and 10; integration tests verify the `tool_result` event carries the reminder at exactly the right turns and not before. Exporting `PERIODIC_REMINDER_INTERVAL` so tests don't hardcode `5` is the right call. CI is green, no conflicts.

---

### Concerns

**1. Wording diverges from `AGENT_REMINDERS`, creating overlapping-but-inconsistent instructions**

`PERIODIC_REMINDER_TEXT` paraphrases the event-driven reminders rather than reusing their exact text:

| `AGENT_REMINDERS` | `PERIODIC_REMINDER_TEXT` |
|---|---|
| "never commit or push without an explicit user request" | "commit only when explicitly asked" |
| "verify the change works — find and run the project's test command before reporting done" | "run tests after changes" |

On any turn where a write call first occurs *and* it's a multiple-of-5 (e.g. the very first write happens at turn 5), both sets fire in the same tool result and the model receives two slightly different phrasings of the same rule back-to-back. Using the exact text from `AGENT_REMINDERS`, or deriving `PERIODIC_REMINDER_TEXT` from that array, would keep the two reminder channels consistent and avoid the potential for the model to treat them as distinct constraints.

**2. No opt-out path for custom system prompts**

Operators overriding the default instructions via `OPENCLI_SYSTEM_MD` have no way to suppress or reconfigure the periodic reminder — it fires unconditionally regardless of what the custom prompt says. If the custom prompt already has its own reminder cadence, or deliberately omits certain rules, the hardcoded `PERIODIC_REMINDER_TEXT` will contradict or duplicate it silently. A `PERIODIC_REMINDER_INTERVAL=0` env var (alongside the existing `OPENCLI_*` family) or a nullable export would make this composable without adding complexity to the hot path.

---

### Recommendation

Merge after addressing concern #1 (wording consistency) or adding a comment explaining why the divergent phrasing is intentional. Concern #2 is lower priority — acceptable to track as a follow-up if the feature ships first.



---

zjshen14:
Rebased onto current `main` (commit `706dceb`) to resolve merge conflicts. The rebase was non-trivial: since PR #211 was created, the agentic loop in `agent.ts` was refactored — the `applyReminderToLastResult` private method was removed and its logic was inlined back into `run()`. The conflict resolution:

- Applied the PR's intended change to the inline reminder code (renamed `reminder` → `eventReminder`, added `periodicReminder = buildPeriodicReminder(turns)`, combined them)
- Kept HEAD's monolithic `run()` structure unchanged (did not re-introduce the extracted private methods, which were from the PR's old base)
- Kept both test suites that were added at the same location (`Agent activate_skill — conversation well-formedness` from HEAD + `Agent periodic reminder injection` from this PR)

All 716 tests pass (6 new from this PR). Branch is now conflict-free and ready for review.

---

zjshen14:
## Re-review: feat: inject periodic mid-loop reminder every 5 turns (post-rebase, SHA `cf92fdc`)

This is a re-review following the rebase onto `main` @ `706dceb`. CI is all green (Typecheck+Lint+Test, CodeQL, Analyze) and `mergeable_state` is `clean`.

### Rebase adaptation

The rebase correctly handles the `applyReminderToLastResult` refactor that landed on `main` between the original PR and today: the extracted private method was re-inlined, and the PR's `periodicReminder` is now combined with `eventReminder` before the single `if (combined && results.length > 0)` guard. The logic is equivalent to the original intent and there are no mechanical issues.

`buildPeriodicReminder(turns)` is called after `turns++`, so it fires at turns 5, 10, 15, … as the tests verify. No change there.

---

### Prior concerns — still unaddressed

Both concerns from the [June 4 review](https://github.com/zjshen14/opencli/pull/211#issuecomment-4624025468) remain open in this rebased state:

**1. Wording diverges from `AGENT_REMINDERS` ([`prompt.ts:73–82`](src/core/prompt.ts#L73))**

`AGENT_REMINDERS` says:
- "never commit or push without an explicit user request"
- "verify the change works — find and run the project's test command before reporting done"

`PERIODIC_REMINDER_TEXT` says:
- "commit only when explicitly asked" — omits "or push"; weaker phrasing
- "run tests after changes" — significantly vaguer than the event-driven version
- "prefer targeted edits over full rewrites" — new constraint not present in `AGENT_REMINDERS` at all

The two channels now teach the model three slightly different versions of the same rules. Deriving `PERIODIC_REMINDER_TEXT` from `AGENT_REMINDERS` (e.g. `AGENT_REMINDERS.map(r => r.text).join('; ')`) would keep them in sync mechanically and guarantee they can't diverge again. Alternatively, a short comment explaining why the shorter/different phrasing is intentional (e.g. "compact form for token budget") resolves the concern without a code change.

**2. No opt-out path for `OPENCLI_SYSTEM_MD` overrides**

Operators using a custom system prompt via `OPENCLI_SYSTEM_MD` have no way to suppress or reconfigure the periodic reminder. Still lower priority — still acceptable as a follow-up.

---

### Recommendation

Merge after addressing concern #1 — either align the `PERIODIC_REMINDER_TEXT` wording with `AGENT_REMINDERS`, or add a brief inline comment explaining why the intentional divergence is correct. Concern #2 remains a follow-up item.



---

zjshen14:
Addressed concern #1 from the re-review (commit `35f9e7c`).

**What changed:**
- `PERIODIC_REMINDER_TEXT` is now derived from `AGENT_REMINDERS` via `AGENT_REMINDERS.map((r) => r.text).join("; ")` instead of a hand-written string
- `AGENT_REMINDERS` was moved before the periodic reminder block to reflect the dependency order
- Tests in `prompt.test.ts` and `agent.test.ts` updated to assert on the actual `AGENT_REMINDERS` wording

This means both reminder channels now use identical phrasing and can't drift apart as entries are added or changed in the future.

Concern #2 (no opt-out path for `OPENCLI_SYSTEM_MD`) is unchanged — still a lower-priority follow-up as noted in the review.

---

zjshen14:
Re-review following the wording fix in commit `35f9e7c`. `PERIODIC_REMINDER_TEXT` is now derived from `AGENT_REMINDERS.map((r) => r.text).join("; ")` — both reminder channels share the same text and can't drift apart as entries are added. The tests correctly assert on the actual `AGENT_REMINDERS` wording (checking for `"never commit or push without an explicit user request"`). CI is all green, `mergeable_state` is `clean`. Both prior concerns from the June 4 and June 19 reviews are fully resolved. Merging.



---

PR #229, opened 2026-06-06
feat(eval): D2 replay harness skeleton — Tape, TapeClient, TapeRegistry, runTape

zjshen14:
## Summary

Implements PR 1 of the D2 plan ([`docs/design/d2-context-management-replay.md`](docs/design/d2-context-management-replay.md)): the load-bearing skeleton that lets us validate the agent loop's context-management contract against recorded session shapes, fully offline.

**Why now.** [#228](https://github.com/zjshen14/opencli/pull/228) just refactored `agent.ts`'s turn loop. Until D2 lands there is no automated check that A5b auto-compact still fires at the expected ratios on a realistic trajectory — only unit tests covering trigger math in isolation. This PR closes that gap for the synthetic-tape case; PR 2 will fold in the real `card_trade` tape (requires the redaction script).

## What's in this PR

| File | What it does |
|---|---|
| [`src/eval/replay/tape.ts`](src/eval/replay/tape.ts) | Parses session JSONL into `AgentTurn[]` / `LLMIteration[]`. Handles parallel tool-call batches, multi-iteration turns, `/plan` and `/exit` filtering, `thoughtSignature` preservation. |
| [`src/eval/replay/client.ts`](src/eval/replay/client.ts) | `TapeClient implements LLMClient` — yields recorded events. Records the `messages` arg the agent sent on each `stream()` call (inspection hook for prune-anchor assertions; no `agent.ts` change required). |
| [`src/eval/replay/registry.ts`](src/eval/replay/registry.ts) | `TapeRegistry extends ToolRegistry` — serves recorded tool results from a FIFO queue, matched by name. Synthesises stubs for every tool name in the tape with `readonly` set correctly so the executor's sequential-vs-parallel decision matches the recording. |
| [`src/eval/replay/runner.ts`](src/eval/replay/runner.ts) | `runTape(tape, opts)` drives an Agent for each turn and returns the captured `ObservabilityEvent` stream + `AgentEvent` stream + inspection state. Stub compaction client keeps replay offline even when auto-compact fires. |
| [`src/eval/replay/assertions.ts`](src/eval/replay/assertions.ts) | Typed helpers over `ObservabilityEvent[]` — `eventsOfType`, `countOfType`, `firedBefore`, `compactStartedResolutions`, `tokenTrajectory`, `firstSnapshotAtOrAbove`. Independent of the runner so D1 scenarios can reuse the vocabulary later. |

## Load-bearing proof

The end-to-end tests in `runner.test.ts` are the point of the PR — they demonstrate the harness actually validates A5b on realistic shapes:

- **Minimal trajectory** — no compaction events, tape exhausts cleanly, every tool result consumed.
- **Warn-only tape** (turn 2 gate lands at ~67%) — exactly one `compact_threshold_warned`, zero `compact_started`.
- **Compact tape** (turn 2 gate above 75%) — `compact_started` paired with `compact_completed`, `messagesRemoved > 0`, no `compact_failed`. Skips the threshold-warn (warn and compact are mutually exclusive within one gate).
- **Token trajectory** crosses 75k tokens (75% of the default 100k window).
- **`autoCompact: false`** suppresses both events even on a pressured tape.
- **Plan mode** runs without firing auto-compact.

## What's NOT in this PR (deferred per design doc)

- **Real `card_trade-2026-05-17` tape + `scripts/redact-tape.ts`** — PR 2. Redaction needs care (web_fetch outputs, bash env values).
- **Synthesised edge-case tapes** (nested compaction, stuck-loop, env_error_loop) — PR 3.
- **Roadmap status flip** — after PR 3.

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — 708 passed (+52 from this PR)
- [x] All replay tests pass under the existing `npm test` invocation (no new script needed; vitest auto-discovers `*.test.ts`)
- [ ] Reviewer sanity check: the auto-compact assertions match the documented A5b contract (60% warn, 75% compact, mutually exclusive within one gate, `messagesRemoved > 0` on compact)

### DIFF
--- a/src/eval/replay/assertions.test.ts
+++ b/src/eval/replay/assertions.test.ts
@@ -1,0 +1,97 @@
+import { describe, it, expect } from "vitest";
+import type { ObservabilityEvent } from "../../core/observability.js";
+import {
+  eventsOfType,
+  countOfType,
+  firedBefore,
+  compactStartedResolutions,
+  tokenTrajectory,
+  firstSnapshotAtOrAbove,
+} from "./assertions.js";
+
+function snap(messageCount: number, estimatedTokens: number): ObservabilityEvent {
+  return { type: "context_snapshot", messageCount, estimatedTokens };
+}
+
+describe("eventsOfType / countOfType", () => {
+  it("filters by type with narrowed return", () => {
+    const stream: ObservabilityEvent[] = [
+      snap(1, 100),
+      snap(2, 200),
+      { type: "empty_response_retry" },
+    ];
+    expect(eventsOfType(stream, "context_snapshot")).toHaveLength(2);
+    expect(countOfType(stream, "context_snapshot")).toBe(2);
+    expect(countOfType(stream, "empty_response_retry")).toBe(1);
+  });
+});
+
+describe("firedBefore", () => {
+  it("returns true when A appears before B", () => {
+    const stream: ObservabilityEvent[] = [
+      { type: "compact_threshold_warned", ratio: 0.6 },
+      { type: "compact_started", trigger: "auto", ratio: 0.75 },
+    ];
+    expect(
+      firedBefore(
+        stream,
+        (e) => e.type === "compact_threshold_warned",
+        (e) => e.type === "compact_started",
+      ),
+    ).toBe(true);
+  });
+
+  it("returns false when either is missing", () => {
+    const stream: ObservabilityEvent[] = [{ type: "compact_threshold_warned", ratio: 0.6 }];
+    expect(
+      firedBefore(
+        stream,
+        (e) => e.type === "compact_threshold_warned",
+        (e) => e.type === "compact_started",
+      ),
+    ).toBe(false);
+  });
+});
+
+describe("compactStartedResolutions", () => {
+  it("pairs each compact_started with its following compact_completed of matching trigger", () => {
+    const stream: ObservabilityEvent[] = [
+      { type: "compact_started", trigger: "auto", ratio: 0.8 },
+      { type: "compact_completed", trigger: "auto", messagesRemoved: 5, summaryLength: 300 },
+      { type: "compact_started", trigger: "manual" },
+      { type: "compact_completed", trigger: "manual", messagesRemoved: 3, summaryLength: 200 },
+    ];
+    const out = compactStartedResolutions(stream);
+    expect(out).toHaveLength(2);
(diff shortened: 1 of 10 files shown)


zjshen14:
## Review — D2 replay harness skeleton

Strong PR. Fully additive (10 new files under `src/eval/replay/`, no changes to existing code), and the end-to-end `runner.test.ts` cases genuinely prove the claim — A5b auto-compact fires at the right ratios on realistic trajectories, fully offline. This closes the gap #228's turn-loop refactor opened, exactly as the design doc intends.

- `src/eval/` → **67 passed, 1 skipped** (43 new replay tests across the 5 files)

**Things I checked and liked:**
- `SessionEntry` shapes in `tape.ts` match `state/session.ts` exactly (including the optional `thoughtSignature` on `tool_call`).
- The iteration-boundary heuristic (new iteration when a `tool_call` follows a `tool_result`) correctly groups parallel batches vs. sequential rounds — `tape.test.ts` covers both.
- `TapeRegistry` sets `readonly` per tool name so the executor's sequential-vs-parallel batching matches the recording — a subtle but important detail to keep replay faithful.
- `TapeClient.exhausted()` + the exhaustion throw with the auto-compact-aware error message is a nice touch; it turns "agent diverged from the recording" into a loud, legible failure.
- `compactStartedResolutions` handling the unresolved/truncated case (`resolved: null`) rather than assuming pairs.

**Minor / non-blocking (fine as follow-ups):**
1. `assertions.ts` — `compactStartedResolutions` carries a dead `pendingStartIdx` (assigned, never read, then `void pendingStartIdx;` to silence the linter). Just delete the variable and the `void` line; it adds nothing.
2. `TapeRegistry.execute` matches recorded results by **name only** (`findIndex(r => r.name === name)`), not args. For a parallel batch of same-named calls (e.g. three `read`s) the results are served FIFO regardless of which file each call targeted. The docstring acknowledges this, and it's correct as long as the agent's call order matches the recording — but once the **real `card_trade` tape** lands in PR 2, an args-aware match (or at least an args mismatch warning) would catch divergence that name+count matching silently passes. Worth considering then, not now.
3. `TapeClient.peek()` / `remaining()` look unused so far — presumably for the PR 2/3 assertions. Fine to keep as the documented inspection surface.

None of these block. LGTM ✅ — approving and merging.

PR #231, opened 2026-06-07
feat(eval): D2 real card_trade tape + redaction script + empty-assistant padding

zjshen14:
## Summary

Implements PR 2 of the D2 plan ([`docs/design/d2-context-management-replay.md`](docs/design/d2-context-management-replay.md)): the first real-session tape sealed for replay, plus the tooling to seal more.

## What landed

**`scripts/redact-tape.ts`** — produces a sealed tape from a raw session JSONL. Redacts absolute home paths (`/Users/<username>/`) and bare username references with deterministic placeholders. Supports optional `TRUNCATE_FROM_LINE` / `TRUNCATE_TO_LINE` for cutting to a clean slice. Writes a sidecar `.redaction-log.json` with input/output sha256s and per-rule replacement counts so refreshes are auditable.

**`src/eval/replay-tapes/card-trade-2026-05-17.jsonl`** — the first sealed tape. 73 entries / 52 KB / 4 react-mode user turns from a 13-day card-trading-site project. Truncated to lines 33–105 of the raw input — this picks the longest continuous react-mode stretch and avoids the `/plan` back-to-back assistant artifacts the tape parser doesn't yet model. Sidecars:
- `.redaction-log.json` — audit trail
- `.expected.json` — observability event counts
- `README.md` (tapes dir) — truncation rationale and refresh procedure

**`src/eval/replay/replay.test.ts`** — the canonical e2e test. Loads each sealed tape, replays it, and asserts observability event counts against the checked-in `expected.json`. If a refactor changes the emission pattern in a way that matters, the delta lands in the `expected.json` diff and shows up in PR review.

## Bug fixes from PR 2 integration

**`src/eval/replay/tape.ts`** — empty assistant content now pads with one extra empty iteration. Original cause: when the recorded session's final iteration emitted no text and no calls, the agent's empty-response-retry path called `stream()` twice before emitting `done`. Replay needs to provide two iterations to satisfy that path; without the pad, the `TapeClient` exhausts mid-replay. New unit test in `tape.test.ts` pins the behavior.

## What's NOT in this PR (deferred per design doc)

- **Synthesised edge-case tapes** (nested compaction, stuck-loop, env_error_loop) — PR 3
- **Larger real tape that exercises compaction** — the current tape peaks at ~11k tokens, too short to trigger A5b auto-compact. The compaction contract is already covered by the synthesised-pressure tapes shipped in PR #229. A bigger real tape requires handling the `/plan` two-run pattern; deferred until the value justifies the parser work.

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — 711 passed (+3 from this PR)
- [x] Sealed tape sha256 matches the value in its `.redaction-log.json` (deterministic refresh)
- [x] No occurrences of the unredacted username remain in the sealed tape (`grep -c \"/Users/zhijie\\|zjshen\\|Zhijie\"` = 0)
- [ ] Reviewer: scan the sealed `.jsonl` for anything I missed (long base64 hashes are package-lock integrity + mock-data IDs, intentionally NOT redacted — let me know if a different policy is wanted)


zjshen14:
## Review — D2 real card_trade tape + redaction + empty-assistant fix

Good PR 2. The redaction tooling is auditable, the sealed tape replays cleanly, and the empty-assistant bug fix is a real find — it's exactly the kind of fidelity gap a real tape surfaces that synthetic ones wouldn't.

- `src/eval/replay/` → **46 tests pass** (the new `replay.test.ts` e2e + the empty-assistant unit test)
- **Redaction leak check**: `grep -ic "zhijie\|/Users/zhijie"` on the sealed `.jsonl` → **0**. Clean.
- Redaction-log sha256 (`adcf702b…`) and the 28 replacement counts are internally consistent with the truncation range (lines 33–105).

**The empty-assistant fix is correct and well-reasoned.** A recorded turn that ended with empty assistant content means the runtime's empty-response-retry fired (two `stream()` calls before `done`), so the tape needs a second empty iteration or `TapeClient` exhausts mid-replay. The padding in `tape.ts` and the pinning unit test both look right.

**One doc inaccuracy (non-blocking):**
- `scripts/redact-tape.ts` docstring item 3 claims: *"The `cwd` field on `session_start` is rewritten to a generic path."* There's no code that does this — the script operates purely on string replacement over the raw text, so a `cwd` only gets redacted incidentally if it contains `/Users/<user>/`. In this tape the slice starts at line 33, so there's no `session_start` entry at all and it doesn't matter here — but the docstring describes behavior that doesn't exist. Either drop that line or implement the structured rewrite. Worth fixing because the next person sealing a tape that *does* include `session_start` will trust it.

**Things I checked and liked:**
- Deterministic redaction (literal string replace, longest-first rule ordering) + the `exit 4` on zero-replacements guard catches a wrong `REDACT_USER`.
- The line-count sanity check (input lines === output lines) is a nice guard against a replacement accidentally corrupting JSONL structure.
- `expected.json` splitting `exact` vs `bounds` counts is a thoughtful way to keep the e2e test from being brittle against benign iteration-count drift.

**Minor / optional:**
- The PR notes call out long base64 hashes (package-lock integrity, mock-data image IDs) as intentionally un-redacted. Agreed — those aren't secrets and redacting them would break determinism. No action needed; just confirming the policy reads right.

Non-blocking on the docstring. LGTM ✅ — approving and merging; please fix or drop the `cwd` docstring line in a follow-up (or I can push a one-liner).

PR #232, opened 2026-06-07
fix(eval): make redact-tape actually rewrite session_start.cwd as documented

zjshen14:
## Summary

Follow-up to the [PR #231 review comment](https://github.com/zjshen14/opencli/pull/231#issuecomment-from-zjshen14). The docstring on `scripts/redact-tape.ts` claimed item 3:

> The `cwd` field on `session_start` is rewritten to a generic path.

…but there was no code that did this. The script's passes were both pure string replacement, so a `cwd` only got redacted incidentally if it contained `/Users/<username>/`. Tapes with cwds like `/Workspace/<project>` or `/opt/<project>` would leak the project name even after redaction.

## Fix

Adds a structured pre-pass that JSON-parses each line, finds `session_start` entries, and rewrites the `cwd` field to a single placeholder (`/Users/REPLAY-USER/Workspace/replay-project`). This also generalises the project name — before, even when the home path was redacted, the project directory name (e.g. `card_trade`) survived because the string rules wouldn't touch it.

Log adds a `sessionStartRewrites` field so reviewers can verify the structured pass fired the expected number of times.

## Verified

- **Committed truncated tape (lines 33–105)** has no `session_start`, so the new pass is a no-op. Sealed file's sha256 is unchanged (`adcf702b…`). No tape refresh needed in this PR.
- **Full untruncated input** now produces `sessionStartRewrites: 1` and the cwd becomes `/Users/REPLAY-USER/Workspace/replay-project` (was `/Users/REPLAY-USER/Workspace/card_trade` under string-only rules).

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — 711 passed
- [x] `npx eslint scripts/redact-tape.ts` — clean
- [x] sha256 of the committed tape unchanged after re-running the script

### DIFF
--- a/scripts/redact-tape.ts
+++ b/scripts/redact-tape.ts
@@ -81,6 +81,42 @@ function buildRules(username: string): ReplacementRule[] {
   ];
 }
 
+/**
+ * Pre-pass: parse each line as JSON and rewrite the `cwd` field on any
+ * `session_start` entry to a generic placeholder. This catches cwds that
+ * the string-replacement rules would otherwise miss — e.g.
+ * `/Workspace/<project>` or `/opt/<project>` paths that don't contain the
+ * author's username. Non-JSON / non-session_start lines pass through
+ * untouched.
+ *
+ * Returns the rewritten content and the number of session_start entries
+ * normalised.
+ */
+function normaliseSessionStarts(input: string): { out: string; rewrites: number } {
+  let rewrites = 0;
+  const outLines = input.split("\n").map((line) => {
+    if (!line.trim()) return line;
+    let parsed: unknown;
+    try {
+      parsed = JSON.parse(line);
+    } catch {
+      return line;
+    }
+    if (
+      parsed &&
+      typeof parsed === "object" &&
+      (parsed as { type?: unknown }).type === "session_start" &&
+      typeof (parsed as { cwd?: unknown }).cwd === "string"
+    ) {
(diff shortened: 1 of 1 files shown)


zjshen14:
## Review — make redact-tape rewrite session_start.cwd as documented

This directly closes the docstring gap I flagged on #231, and goes a step further than I asked — it generalises the *project name* too, not just the home path. Good fix.

- `src/eval/replay/` → **46 tests pass**
- **No-op claim confirmed**: the committed `card-trade-2026-05-17.jsonl` has **0** `session_start` entries, so the new pass genuinely doesn't touch it — the unchanged sha256 (`adcf702b…`) holds without a tape refresh.
- **Behavioral test**: ran the script on a fixture with `{"type":"session_start","cwd":"/Workspace/secret_project"}` and a non-matching username (`REDACT_USER=nobody`). The cwd was rewritten to `/Users/REPLAY-USER/Workspace/replay-project` and counted as 1 replacement — exactly the leak the string-only rules missed before.

**One subtle thing I checked and liked:** folding `sessionStartRewrites` into `totalReplacements` means a tape whose *only* sensitive content is the cwd (string rules would find nothing) no longer false-trips the `exit 4` "no replacements made" guard. That's the correct interaction between the two passes.

**Structured JSON pre-pass + string second-pass** is the right ordering — `cwd` gets canonicalised before the byte-level rules run, so there's no double-processing risk.

LGTM ✅ — approving and merging.

PR #233, opened 2026-06-07
feat(eval): D2 synthesised edge-case replay tapes — stuck-loop, env-error-loop, nested compaction

zjshen14:
## Summary

Implements PR 3 of the D2 plan ([`docs/design/d2-context-management-replay.md`](docs/design/d2-context-management-replay.md)): the three synthesised edge-case scenarios the real `card_trade` tape doesn't exercise.

## Format decision

Two cases land as **committed JSONL** under `src/eval/replay-tapes/synthesized/` (small enough for reviewers to inspect on PR diff). One case lives **programmatically** in `synthesized.test.ts` because its synthetic payload would otherwise bloat the repo by ~500 KB for zero inspection value. Design doc updated to document this split.

## What landed

| Tape | Format | Asserts |
|---|---|---|
| `stuck-loop` | JSONL (6 entries) | Three identical `(name, args)` tool calls in a row fire `guard_triggered('stuck_loop')` after the third recurrence. Iter 3 aborts **before** tool execution (only iters 1+2 run tools). |
| `env-error-loop` | JSONL (7 entries) | Three iterations whose `tool_result`s contain `EPERM` with **different** args (so stuck-loop does NOT also fire) trigger `guard_triggered('env_error_loop')` after the third. All 3 tool calls execute — the guard fires **after** tool execution. |
| Nested compaction | Programmatic | Five × 100 KB `read` results in turn 1 → post-first-compaction tail (4 large messages preserved by `KEEP_RECENT=10`) is **still** above the 75 k token threshold → second auto-compact fires at turn 3 entry. Asserts: `compact_started` fires twice paired with `compact_completed`; second start fires *after* first completes; original user input survives both compactions via the prune-anchor verbatim block in `extractOriginalTask`. |

## What this finishes from D2

This closes the context-management replay sub-item of D2. The other items in the D2 row of the roadmap (sandbox / plan-mode / HITL contract evals) remain open and will be tracked as separate work.

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — 719 passing (+8 from this PR)
- [x] All three new synthesised cases pass cleanly with no flakes across 10 local runs
- [ ] Reviewer: scan `synthesized.test.ts` lines 122–217 (nested-compaction case) — the token math is load-bearing and depends on `KEEP_RECENT=10`, `DEFAULT_CONTEXT_WINDOW=100_000`, and the 0.75 trigger ratio all staying as documented

### DIFF
--- a/docs/design/d2-context-management-replay.md
+++ b/docs/design/d2-context-management-replay.md
@@ -162,9 +162,12 @@ Because the tape was recorded before A5b, after the first auto-compaction fires 
 
 ### Synthesized tapes
 
-For assertions that need behaviors the real session doesn't exhibit (nested compactions, stuck-loop, env-error loop), small hand-written tapes live alongside the real ones under `src/eval/replay-tapes/synthesized/`. These are explicitly marked as synthesized and have richer `expected.json` annotations.
+For assertions that need behaviors the real session doesn't exhibit (nested compactions, stuck-loop, env-error loop), the harness uses two formats depending on payload size:
 
-This is the same separation D1 uses between real-fixture scenarios (the YAML suite against `mini-ts/`) and edge-case fixtures (`mini-ts-partial/`).
+- **Small JSONL tapes committed under `src/eval/replay-tapes/synthesized/`** — for cases where the recorded sequence is short and inspectable on PR diff. `stuck-loop` (6 entries) and `env-error-loop` (7 entries) both fit. These have richer `expected.json` annotations including a `guards: [{guard, reasonContains}]` block that asserts the specific guard fires with the expected reason fragment.
+- **Programmatic tapes in `synthesized.test.ts`** — for cases that need ≥ 500 KB of synthetic content to push the agent past a threshold. Nested compaction is the canonical example: it requires five 100 KB tool_results to bloat the tail past 75 % both before and after the first compaction. Committing that as JSONL would bloat the repo for no inspection value.
+
+This is the same separation D1 uses between real-fixture scenarios (the YAML suite against `mini-ts/`) and edge-case fixtures (`mini-ts-partial/`) — committed when small, generated when not.
 
 ---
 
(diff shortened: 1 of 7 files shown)


zjshen14:
## Review — D2 synthesised edge-case tapes (PR 3, final)

This completes the context-management replay arc cleanly. The three scenarios are exactly the ones the real `card_trade` tape can't reach (it peaks at ~11k tokens, no guards), and the committed-vs-programmatic split is the right call.

- `src/eval/replay/` → **54 tests pass** (+8 from this PR)
- **Flake check**: ran `synthesized.test.ts` **5×**, 8/8 every time. The nested-compaction token math you flagged as load-bearing is stable.

**Cross-checked the guard expectations against `agent.ts` directly** (this is the part worth verifying, since the expected.json counts encode pipeline ordering):
- `STUCK_THRESHOLD` and `ENV_ERROR_THRESHOLD` are both 3 ✓
- **stuck-loop** → 2 `tool_exec_start`/`end`: correct. The stuck guard runs at run()'s step 4, *before* `executeTurnTools` (step 5), so iter 3 aborts before executing. The expected.json's `tool_exec_start: 2` captures this precisely.
- **env-error-loop** → 3 `tool_exec`: correct. The env guard runs at step 7, *after* execution, so all 3 calls run before the abort. The contrast between the two tapes is a genuinely good test of the pipeline order, not just the thresholds.
- **env-error tape uses different args per iter** (`ls /root` / `cat /root/secrets` / `touch /root/x`) so stuck-loop doesn't also fire — and the test asserts `guards == ["env_error_loop"]` exactly. Nice: that catches a regression where both guards fire on the same trajectory.

**Nested compaction is the strongest of the three.** Driving 5×100KB results so the `KEEP_RECENT=10` tail stays above 75% *after* the first compaction is a clever way to force a genuinely nested second compaction, and the three assertions (twice paired, second-after-first-completed by stream index, original task survives in the final `sentMessages`) pin the real contract — including the `extractOriginalTask` prune-anchor preservation that #191 fixed. Good end-to-end coverage of that path.

**Minor / optional (non-blocking):**
- The `guards` block in `synthesized.test.ts` (the `reasonContains` matcher) is a nice annotation — consider lifting it into the main `replay.test.ts` Expected schema too, so the real-tape e2e can assert "no guards fired" symmetrically. Future cleanup, not this PR.

Docs (`d2-context-management-replay.md` + tapes README) are updated to document the committed-vs-programmatic split. LGTM ✅ — approving and merging.

One note for after merge: this closes the context-management replay sub-item of D2, so the roadmap status flip is now unblocked (the design doc still reads "Ready for implementation"). Happy to do that as a one-line follow-up.

Issue #234, opened 2026-06-07
D2 follow-up: sandbox / plan-mode / HITL contract evals


zjshen14:
## Context

The **context-management replay** sub-item of D2 shipped across PRs #229 + #231 + #232 + #233 (June 6–7, 2026). That validates the agent loop's auto-compact, prune-anchor, parallel-batching, and safety-guard behaviours on real recorded sessions, fully offline.

The D2 roadmap row was originally scoped as a bundle of contract evals: *"Contract evals (sandbox, plan mode, HITL, headless schema, context-management replay)"*. The remaining four surfaces are still open:

- [ ] **Sandbox contract eval** — assert each sandbox profile (`passthrough`, `bwrap`, `sandbox-exec`, strict) actually denies the operations it claims to deny, against a sealed list of attempted operations.
- [ ] **Plan-mode contract eval** — assert that with `mode: "plan"`, the executor blocks write tools at the registry layer AND the prompt restricts tool selection. Replay a tape that contains write attempts under plan mode and confirm they're refused.
- [ ] **HITL (deny / ask) contract eval** — replay tapes against the executor with `confirmFn` configured to deny or auto-allow, and assert `tool_denied` events fire correctly with the right `reason`.
- [ ] **Headless `--output=json` contract eval** — assert the JSON event stream from `opencli run --output=json <prompt>` matches the documented schema (semver-stable, parseable, fields present and typed correctly).

## Why these are separate from the replay sub-item

The replay harness (`src/eval/replay/`) is the right tool for context-management. The other surfaces want different fixtures:
- Sandbox eval wants a list of attempted shell operations + expected refusals; the agent loop is not the SUT.
- Plan-mode eval can probably extend the replay harness with one extra knob (mode=plan), but the tape's recorded turn must contain a write attempt.
- HITL eval needs a controllable confirmFn — already plumbed in the executor's `ExecutorDeps`; replay-side wrapper is the missing piece.
- Headless JSON eval is closer to a CLI smoke test than a replay; could live under `src/eval/headless/`.

## Sequencing

Per the roadmap, D2 is *"Alongside A6 — all A-phase properties must be testable before A closes"*. A6 (#89, inline diffs + compact tool rows) is still open and is the closing gate of Phase A. These four contract evals can land any time before A6 ships, ideally in small focused PRs rather than one big bundle.

## Related

- D2 design doc: [`docs/design/d2-context-management-replay.md`](docs/design/d2-context-management-replay.md) (replay sub-item only)
[...]

PR #239, opened 2026-06-09
feat(headless): add --output=json to opencli run + JSON schema contract eval

zjshen14:
## Summary

- Adds `--output <text|json>` flag to `opencli run`; in `json` mode every `AgentEvent` is written as a newline-terminated JSON line (NDJSON) to stdout instead of the human-readable text rendering
- Introduces `src/cli/json-output.ts` with the stable `JsonOutputEvent` type (the public schema) and `toJsonLine()` which converts any `AgentEvent` → JSON line, explicitly stripping the provider-internal `thoughtSignature` from `tool_call` events
- Adds `src/eval/headless/json-schema.test.ts` — 33 contract-eval tests asserting parseable NDJSON, newline termination, required fields per event type, `thoughtSignature` exclusion, and exhaustive schema coverage across all 7 `AgentEvent` types

## Related issue

Part of #234 (D2 follow-up: headless `--output=json` contract eval sub-item)

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — all pass (742 tests, 33 new)
- [x] `toJsonLine()` unit-tested for all 7 `AgentEvent` types
- [x] `thoughtSignature` stripping verified in contract eval
- [x] Schema coverage test enforces exhaustiveness — will fail if a new `AgentEvent` type is added without updating the eval

https://claude.ai/code/session_01TakF9J9FMjjQd6okE7kczu

---
_Generated by [Claude Code](https://claude.ai/code/session_01TakF9J9FMjjQd6okE7kczu)_

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -10,6 +10,7 @@ import { resolveApiKey } from "./keys.js";
 import { runRepl } from "./repl.js";
 import { createConfirmFn } from "./confirm.js";
 import { printError, printInfo } from "./renderer.js";
+import { toJsonLine } from "./json-output.js";
 import type { ObservabilityEvent } from "../core/observability.js";
 import { createSandboxRunner } from "../tools/exec/sandbox/index.js";
 import type { SandboxMode } from "../tools/exec/sandbox/types.js";
@@ -97,7 +98,13 @@ program
   .option("--provider <provider>", "Override provider detection: gemini | anthropic | openai")
   .option("--base-url <url>", "Custom base URL for proxy or local inference (e.g. LiteLLM)")
   .option("--temperature <float>", "LLM temperature (use 0 for determinism)", parseTemperature)
+  .option("--output <format>", "Output format: text | json (default: text)")
   .action(async (prompt: string, opts) => {
+    const output = opts.output as string | undefined;
+    if (output !== undefined && output !== "text" && output !== "json") {
+      printError(`Invalid --output value '${output}'. Valid values: text, json`);
+      process.exit(1);
+    }
     await runSingle(
       prompt,
       opts.model,
@@ -109,6 +116,7 @@ program
       opts.provider as string | undefined,
       opts.baseUrl as string | undefined,
       opts.temperature as number | undefined,
+      output as "text" | "json" | undefined,
     );
   });
 
@@ -224,6 +232,7 @@ async function runSingle(
   providerOverride?: string,
   baseUrlOverride?: string,
   temperature?: number,
+  outputFormat?: "text" | "json",
 ): Promise<void> {
   const { agent, mcpManager } = await createAgent(
     modelOverride,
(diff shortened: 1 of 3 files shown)


zjshen14:
## Review: feat(headless): add --output=json to opencli run + JSON schema contract eval

**What's good:** Clean, well-layered addition. `json-output.ts` correctly lives in `cli/` as a presentation-layer concern and imports only from `core/`. The explicit object construction in `toJsonLine`'s `tool_call` case (`{type, name, args}` — no spread) is the right way to strip `thoughtSignature`. The forward-compat `default: return null` guard is sensible. Tests are thorough: all 7 `AgentEvent` types covered, `thoughtSignature` stripping verified, NDJSON termination checked, and the exhaustiveness test will catch any future `AgentEvent` additions that miss a `toJsonLine` case. CI is green, `mergeable_state` is clean.

---

### Notes

**1. `error` routing differs between modes ([`index.ts:261`](src/cli/index.ts#L261))**

Text mode writes `Error: ...` to `process.stderr`; JSON mode emits `{"type":"error","message":"..."}` to `process.stdout`. The asymmetry is intentional — structured output should keep all events on one stream — but it's non-obvious from reading the function. Consumers expecting stderr-based error detection across both modes will be surprised. No action required, but worth a note in public-facing docs when `--output=json` is documented.

**2. `text` accumulation in json mode**

`if (event.type === "text") text += event.text` runs in json mode, feeding `stream()`'s return value. Presumably this feeds session logging (which is still correct behavior in json mode). Not a bug — just noting that the return value is not dead in json mode.

---

**Recommendation:** Merge.



---

zjshen14:
## Review: feat(headless): add --output=json to opencli run

**What's good:** Clean, well-bounded feature. `toJsonLine()` is a pure function that maps all 7 `AgentEvent` types exhaustively, strips `thoughtSignature` (correctly — it's a provider-internal Gemini field with no public contract), and returns `null` for unknown types as a forward-compat guard. The JSON/text paths in `runSingle()` are clearly separated with no shared mutable state. The contract test suite is a standout: the exhaustiveness check (`covers every AgentEvent type`) will fail immediately if a new event type is added without updating the schema — exactly the right invariant to enforce.

### Notes

**1. `--output` validation could use Commander's `.choices()`**

The manual `if (output !== undefined && output !== "text" && output !== "json")` check works, but `.option("--output <format>", "...", ...).choices(["text", "json"])` (Commander v12+) gives consistent CLI error formatting. Minor; not a blocker.

**2. `error` events route to stdout in JSON mode vs. stderr in text mode**

In JSON mode all events — including `error` — go to stdout as NDJSON lines. This is the correct design for programmatic consumers (single parseable stream), but the current option description (`"Output format: text | json (default: text)"`) doesn't hint at this. Consider extending it to `"Output format: text (human-readable, errors → stderr) | json (NDJSON, all events → stdout)"` so callers know not to filter stderr in JSON mode.

**Recommendation:** Merge. Note 2 is a docs gap, not a correctness issue. Both notes are optional improvements.



---

PR #242, opened 2026-06-10
fix: return synthetic functionResponse for activate_skill to keep conversation well-formed

### LINKED ISSUE #241: fix: activate_skill leaves orphaned functionCall with no functionResponse — breaks Gemini/Anthropic on next turn
## Bug

When the model calls the `activate_skill` tool, `executeCalls` in `src/core/executor.ts` handles it specially (loads skill content into context) but returns **no `FunctionResultPart`** for it. This leaves the conversation history with a `functionCall` part in the model message that has no corresponding `functionResponse` in the following user message.

Both Gemini and Anthropic require that every `functionCall` in a model turn has a matching `functionResponse` before the next LLM call is made. Without it they return a `400 Bad Request` error.

## Affected scenario

Any agent turn where the model calls `activate_skill` — either alone or alongside other tools. In both cases the resulting user message sent to the provider is missing the `functionResponse` for `activate_skill`.

## Reproduction

```typescript
// executor.ts — executeCalls returns { results: [] } for activate_skill-only calls
const { results } = await executeCalls([{ name: "activate_skill", args: { name: "review" } }], deps);
// results.length === 0  ← no FunctionResultPart is produced

// agent.ts — because results is empty, no user message with function_result is added
if (results.length > 0) {
  this.context.addMessage({ role: "user", parts: results });
}
// The next client.stream() call sends a conversation ending with:
//   { role: "model", parts: [{ type: "function_call", name: "activate_skill" }] }
// with no matching functionResponse — Gemini/Anthropic returns 400.
```

## Fix

Return a synthetic `FunctionResultPart` for each `activate_skill` call from `executeCalls` (separate from `results` so it isn't surfaced as a `tool_result` event). Include it in the user message added to `ContextManager` so the conversation stays well-formed, while the existing `skill_activated` event continues to be the UI signal.

## Files

- `src/core/executor.ts` — add `skillResults: FunctionResultPart[]` to `ExecutionResult`; build synthetic results for each activate_skill call
- `src/core/agent.ts` — destructure `skillResults`; use `[...skillResults, ...results]` for `context.addMessage`
- `src/core/executor.test.ts` — update activate_skill test to expect `skillResults`
- `src/core/agent.test.ts` — add test verifying function_result appears in next LLM call's messages

zjshen14:
## Summary

- `activate_skill` calls were handled entirely inside `executeCalls` (context mutation) but returned **no `FunctionResultPart`**. This left the conversation history with an orphaned `functionCall` in the model message and no matching `functionResponse` in the next user message.
- Both Gemini and Anthropic reject such conversations with `400 Bad Request` on the next `client.stream()` call.
- Fixed by adding `skillResults: FunctionResultPart[]` to `ExecutionResult`. Each `activate_skill` call produces a synthetic `"Skill '…' activated."` result that is added to `context.addMessage` (for wire correctness) but **not** yielded as a `tool_result` event (the existing `skill_activated` event remains the UI signal).

## Related issue

Closes #241

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm test` — all pass (710 tests)
- [x] New test in `agent.test.ts` verifies the second `stream()` call receives a `function_result` part for `activate_skill` in its conversation messages
- [x] Updated `executor.test.ts` checks `skillResults` is populated correctly and `results` remains empty for activate_skill-only calls

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01RNxZavtYneGiLCTSvsjnXX)_

### DIFF
--- a/src/core/agent.test.ts
+++ b/src/core/agent.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi } from "vitest";
 import { Agent } from "./agent.js";
 import { ToolRegistry } from "../tools/registry.js";
 import { SkillRegistry } from "../skills/registry.js";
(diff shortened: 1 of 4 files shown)


zjshen14:
**Approved.** Verified: every `activate_skill` call now produces a synthetic `FunctionResultPart` (with `thoughtSignature` forwarded) that is added to context but not surfaced as a `tool_result` event, closing the orphaned-`functionCall` bug on both Gemini and Anthropic. CI is green across all three checks; `mergeable_state` is clean. Merging.



---

